PR #14, opened 2022-07-29
chore: relicense the project as Apache-2.0

MylesBorins:
This is to align with the preferred license of the SigStore Project

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,202 @@
-MIT License
 
-Copyright (c) 2022 GitHub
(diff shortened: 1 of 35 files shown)


feelepxyz (approved):
Thanks for doing the work here @MylesBorins 🙇  This looks good to me. Slightly annoying to add the license header to each file but seems the other Sigstore projects also do this so 🤷

bdehamer (approved):
Thanks @MylesBorins!

PR #15, opened 2022-07-29
support for verifying merkle inclusion proofs

bdehamer:
Adds a `verifyInclusion` function which can validate the Merkle tree inclusion proof information embedded in a Rekor bundle.

Probably want to integrate this into the `Verifier` class so that the inclusion proof can be checked anytime we retrieve a bundle from Rekor.

Largely inspired by https://github.com/transparency-dev/merkle/blob/main/proof/verify.go#L46

### DIFF
--- a/src/merkle/verify.ts
+++ b/src/merkle/verify.ts
@@ -1,0 +1,110 @@
+import { Hasher } from './digest';
+/*
+Copyright 2022 GitHub, Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src/merkle/verify.ts
```diff
@@ -0,0 +1,110 @@
...
+  return proof.reduce((acc, h) => hasher.hashChildren(h, acc), seed);
+}
+
+function innerProofSize(index: bigint, size: bigint): number {
+  return (index ^ (size - BigInt(1))).toString(2).length;
+}
+
+// Counts the number of ones in the binary representation of the given number.
+// https://en.wikipedia.org/wiki/Hamming_weight
+function onesCount(x: bigint): number {
```
kommendorkapten, line 108: A common name for this function is popount (population count)
https://vaibhavsagar.com/blog/2019/09/08/popcount/

feelepxyz (dismissed):
This looks great overall and seeing that this logic is mostly copied from the merkle project gives me confidence the logic is probably pretty solid 💯 

I'm not adding much with this review beyond checking tests, naming and formatting 😄

eddiezane (approved):

PR #16, opened 2022-08-01
Fix oauth cross platform browser open

### LINKED ISSUE #17: Trying to follow example
**Description**

I was following the steps in the README.md but when I try to sign I just get:

```
[midawson@midawson sigstore-js]$ ./bin/sigstore.js sign ./node-addon-api-4.3.0.tgz >signature
Error: No identity token provided
    at Signer.getIdentityToken (/home/midawson/sigstore/sigstore-js/dist/sign.js:59:15)
    at async Signer.sign (/home/midawson/sigstore/sigstore-js/dist/sign.js:21:31)
    at async sign (/home/midawson/sigstore/sigstore-js/dist/cli/index.js:48:23)
    at async cli (/home/midawson/sigstore/sigstore-js/dist/cli/index.js:27:13)
    at async Object.processArgv (/home/midawson/sigstore/sigstore-js/dist/cli/index.js:79:9)
```

I'm running on RHEL 8.  Is there some additional configurationt that I need to get the brower window to pop up ?

**Version**

built from main on  Aug 3 2022

eddiezane:
This PR adds support for opening URLs in Linux and Windows.

Fixes #17

### DIFF
--- a/src/identity/oauth.test.ts
+++ b/src/identity/oauth.test.ts
@@ -71,6 +71,22 @@ describe('OAuthProvider', () => {
 
       expect(token).toBe(idToken);
     });
+
+    it('errors on unsupported platforms', async () => {
+      const platform = Object.getOwnPropertyDescriptor(process, 'platform');
+
+      Object.defineProperty(process, 'platform', { value: 'fakeos' });
+
+      await expect(subject.getToken()).rejects.toBe(
+        'OAuth: unsupported platform: fakeos'
+      );
+
+      Object.defineProperty(
+        process,
+        'platform',
+        platform as PropertyDescriptor
+      );
+    });
   });
 });
 
--- a/src/identity/oauth.ts
+++ b/src/identity/oauth.ts
@@ -110,10 +110,16 @@ export class OAuthProvider implements Provider {
       });
     });
 
-    // Open browser to start authorization request
-    const authBaseURL = await this.issuer.authEndpoint();
-    const authURL = this.getAuthRequestURL(authBaseURL);
-    await this.openURL(authURL);
+    try {
+      // Open browser to start authorization request
+      const authBaseURL = await this.issuer.authEndpoint();
+      const authURL = this.getAuthRequestURL(authBaseURL);
+      await this.openURL(authURL);
+    } catch (err) {
+      // Prevent leaked server handler on error
+      server.close();
+      throw err;
+    }
 
     return result;
   }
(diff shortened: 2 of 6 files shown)


src/identity/oauth.ts
```diff
@@ -113,6 +113,8 @@ export class OAuthProvider implements Provider {
     // Open browser to start authorization request
     const authBaseURL = await this.issuer.authEndpoint();
     const authURL = this.getAuthRequestURL(authBaseURL);
+    // DO NOT SUBMIT
+    // TOOD: This failure is swallowed. How do we raise this?
     await this.openURL(authURL);
```
eddiezane, line 118: This rejected promise is swallowed and I am not sure where to throw it.

src/identity/oauth.ts
```diff
@@ -175,10 +177,26 @@ export class OAuthProvider implements Provider {
...
+      switch (process.platform) {
+        case 'darwin':
+          open = 'open';
+          break;
+        case 'linux':
+          open = 'xdg-open';
+          break;
+        case 'win32':
+          open = 'start';
+          command = `"" ${command}`
```
eddiezane, line 193: Need to confirm this in a Windows vm.

eddiezane: Confirmed this works.

src/identity/oauth.ts
bdehamer: I think this may be as simple as removing the `.catch` on this line:

https://github.com/github/sigstore-js/blob/main/src/sign.ts#L111

This is swallowing any errors which are thrown from within an identity provider. This code has been through a lot of refactoring and it may have made sense to swallow these errors in some past iteration, but I don't think it does anymore.

src/identity/oauth.ts
```diff
@@ -175,10 +181,26 @@ export class OAuthProvider implements Provider {
...
+          open = 'xdg-open';
+          break;
+        case 'win32':
+          open = 'start';
+          command = `"" ${command}`;
+          break;
+        default:
+          return reject(`OAuth: unsupported platform: ${process.platform}`);
+      }
+      console.log(`Your browser will now be opened to: ${url}`);
```
bdehamer, line 202: I think we might want to send this to stderr w/ `console.error` (similar to how it's done in [cosign](https://github.com/sigstore/sigstore/blob/ca3b544f985f87114485ffa686dd558f80357449/pkg/oauth/oidc/interactive.go#L186)). That way if you're redirecting the output of the command to a file to capture the signature/attestation you won't also get this string.

src/identity/oauth.test.ts
```diff
@@ -36,6 +36,12 @@ describe('OAuthProvider', () => {
     mockedExec.mockReset();
   });
 
+  const mockedConsoleLog = jest.spyOn(console, 'log').mockImplementation();
+
+  afterAll(() => {
+    mockedConsoleLog.mockRestore();
+  });
```
bdehamer, line 43: If we send the "Your browser will open..." message to stderr we can ditch this bit -- I've already got `console.error` stubbed out [here](https://github.com/sigstore/sigstore-js/blob/main/jest.setup.js#L16).

bdehamer (changes requested):
Looks good. Just one small request wrt to the "Your browser will open..." message. Thanks!

src/identity/oauth.ts
eddiezane: Great catch!

bdehamer (approved):
Nice!

cpanato (approved):
approving because was already approve by a mantainer, just to unblock and get this merged, was just a rebase

PR #35, opened 2022-08-10
use git hashes instead of tags for actions

cpanato:
#### Summary
- use git hashes instead of tags for actions
Across sigstore org we use the git hash instead of the git tags for the actions to pin in a specific git commit


bdehamer (approved):
👍

PR #39, opened 2022-08-11
make certificate mandatory for verification

### LINKED ISSUE #36: Signing certificate should be mandatory for verification workflow
**Description**

The current verification workflow accept an optional signing certificate parameter. If the certificate is not provided, it is looked-up in Rekor by searching for an entry with a matching hash. This is not recommended -- the certificate should be a required parameter for the verification workflow.

bdehamer:
Closes #36 

**Summary**
Updates the `sigstore.verify` function to make the signing certificate a required parameter. Previously, we were searching the Rekor log to retrieve the signing certificate, but this is not the recommended approach.

There is a lot more to be done as part of the verify workflow -- currently, there is verification against the Rekor log at all (either offline or online). We'll come back and address this in a separate PR.

**Release Notes**
* `sigstore.verify` now requires the signing certificate be passed as a parameter
* `dsse.verify` now requires the signing certificate be passed as a parameter
* `Verifier.verify` no longer searches Rekor to find the signing certificate
* The `sign` CLI command now saves the Fulcio-issued signing certificate so it can be saved for verification purposes
* The `verify` CLI command now requires the path to a file containing the signing certificate be supplied
* README update with new instructions for using `sign` and `verify`.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -15,6 +15,7 @@ limitations under the License.
 */
 import fs from 'fs';
 import { sigstore, dsse } from '../index';
+import { base64Decode } from '../util';
 
 async function cli(args: string[]) {
   switch (args[0]) {
@@ -25,10 +26,10 @@ async function cli(args: string[]) {
       await signDSSE(args[1], args[2]);
       break;
     case 'verify':
-      await verify(args[1], args[2]);
+      await verify(args[1], args[2], args[3]);
       break;
     case 'verify-dsse':
-      await verifyDSSE(args[1]);
+      await verifyDSSE(args[1], args[2]);
       break;
     default:
       throw 'Unknown command';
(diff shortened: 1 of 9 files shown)


src/cli/index.ts
```diff
@@ -43,6 +44,8 @@ const signOptions = {
 async function sign(artifactPath: string) {
   const buffer = fs.readFileSync(artifactPath);
   const signature = await sigstore.sign(buffer, signOptions);
+  const cert = base64Decode(signature.cert);
+  await fs.writeFileSync('signingcert.pem', cert, { flag: 'wx' });
```
bdehamer, line 48: At some point we'll need a *real* CLI -- everything in here is so hacky 😣

feelepxyz (approved):
Nice one! LGTM 👍

Issue #32, opened 2022-08-09
Set user agent


eddiezane:
Set the user agent to `sigstore-js-<version>` for usage monitoring.

bdehamer:
Closed by https://github.com/sigstore/sigstore-js/pull/52

PR #52, opened 2022-08-16
added the user agent to req header

ejahnGithub:

#### Summary
Add user agent in request header, so the server side will know the request is from sigstroe-js. 

Issue: https://github.com/sigstore/sigstore-js/issues/32

### DIFF
--- a/bin/sigstore.js
+++ b/bin/sigstore.js
@@ -14,4 +14,4 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-require('../dist/cli').processArgv();
+require('../dist/src/cli').processArgv();
--- a/bin/sigstore.js
+++ b/bin/sigstore.js
@@ -14,4 +14,4 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-require('../dist/src/cli').processArgv();
+require('../dist/cli').processArgv();
--- a/src/util.ts
+++ b/src/util.ts
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import os from 'os';
-import packageJsonData from '../package.json';
 
 const BASE64_ENCODING = 'base64';
 const UTF8_ENCODING = 'utf-8';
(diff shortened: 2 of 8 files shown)


tsconfig.json
```diff
@@ -1,11 +1,13 @@
...
   "extends": "@tsconfig/node12/tsconfig.json",
   "compilerOptions": {
+    "resolveJsonModule": true,
     "noImplicitAny": true,
     "outDir": "./dist"
   },
   "include": [
-    "src/**/*"
+    "src/**/*",
+    "package.json"
```
bdehamer, line 10: I'm not sure this is necessary, I believe that the "package.json" will always be included by default.

bdehamer (changes requested):
This is a great first pass. I've got a couple of clean-up items in the comments below. Also, let's be sure to add a unit test for the `getUserAgent` function. We should also be able to update some of the existing tests for the Fulctio/Rekor clients to ensure that the new user agent header is being included.

tsconfig.json
ejahnGithub: solved

src/util.ts
```diff
@@ -55,3 +58,39 @@ export const promiseAny = async <T>(
...
+    case 'openbsd':
+      return 'OpenBSD platform';
+    case 'sunos':
+      return 'SunOS platform';
+    case 'win32':
+      return 'windows platform';
+    default:
+      return 'unknown platform';
+  }
+};
```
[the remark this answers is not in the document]
ejahnGithub: solved

bin/sigstore.js
```diff
@@ -14,4 +14,4 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-require('../dist/cli').processArgv();
+require('../dist/src/cli').processArgv();
```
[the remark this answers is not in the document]
ejahnGithub: umm. i got this error message when i am building 
```
node:internal/modules/cjs/loader:959
  throw err;
  ^

Error: Cannot find module '../dist/cli'
```

src/util.ts
```diff
@@ -55,3 +58,10 @@ export const promiseAny = async <T>(
     (value) => Promise.resolve<T>(value)
   );
 };
+
+// Format User-Agent: <product> / <product-version> (<platform>)
+// source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent
+export const getUserAgent = (): string => {
+  const platformName = os.platform();
+  return `sigstore-js/${packageJson.version} (${platformName})`;
```
feelepxyz, line 66: Thoughts on also including the node version and architecture alongside the platform? Not sure it needs to be blocking, could be a follow up maybe?

feelepxyz: not sure if we need to account for non-node envs like deno

ejahnGithub: sounds good! i can add it as my next pr :)

bdehamer (approved):
Nicely done!

PR #55, opened 2022-08-19
added node version in user agent

ejahnGithub:

#### Summary
small pr for adding node version in user agent

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -54,5 +54,6 @@ export const getUserAgent = (): string => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const packageVersion = require('../package.json').version;
   const platformName = os.platform();
-  return `sigstore-js/${packageVersion} (${platformName})`;
+  const nodeVersion = process.version;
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -54,5 +54,6 @@ export const getUserAgent = (): string => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const packageVersion = require('../package.json').version;
   const platformName = os.platform();
-  return `sigstore-js/${packageVersion} (${platformName})`;
+  const nodeVersion = process.version;
+  return `sigstore-js/${packageVersion} (Node ${nodeVersion}) (${platformName})`;
```
bdehamer, line 58: Since we're in here making changes, let's add the `process.arch` value as well. So we'll end-up with something like:
```
sigstore-js/0.0.0 (Node v16.15.0) (darwin/arm64)
```

bdehamer (changes requested):
Looks good, but let's add one more thing

bdehamer (approved):
Nice!

PR #61, opened 2022-08-24
added search entry log and tests

ejahnGithub:

#### Description 

The Rekor client has a new searchLog method which invokes the API linked above.
For now, can limit look-ups to just HashedRekordKind and IntotoKind entries.
New TypeScript interfaces created in support of API
Unit tests implemented (w/ mocked API)


src/client/rekor.ts
```diff
@@ -115,9 +115,9 @@ export interface SearchIndex {
 }
 
 export interface SearchLogQuery {
-  entries?: HashedRekordOptions[];
+  entries?: EntryKind[];
   entryUUIDs?: string[];
-  logIndexes: number[];
+  logIndexes?: number[];
 }
```
bdehamer, line 121: At some point I guess I had started to add this API myself. Have no memory of this 😆

src/client/rekor.ts
```diff
@@ -236,6 +236,25 @@ export class Rekor {
...
+
+    const response = await this.fetch(url, {
+      method: 'POST',
+      body: JSON.stringify(opts),
+      headers: { 'Content-Type': 'application/json' },
+    });
+    checkStatus(response);
+
+    const data = await response.json();
+    return data;
```
bdehamer, line 256: Note that the `Entry` type doesn't quite match the response format that comes back from the API. Instead of returning the raw response you'll want to map the results through `entryFromResponse` so they get translated into the correct format.

The reason for the mismatch is to work around the weird format Rekor uses where the entry is keyed by UUID. Instead of:

```json
{
  "5c1419ace1915869eef4426701d0763232da44d453e31ffc06d60ea61de36e25": {
    "body": "abc",
    "integratedTime": 103930923,
    . . . 
}
```
We translate it to:
```json
{
  "uuid": "5c1419ace1915869eef4426701d0763232da44d453e31ffc06d60ea61de36e25",
  "body": "abc",
  "integratedTime": 103930923,
  . . .
}
```

src/client/tests/constants.ts
```diff
@@ -0,0 +1,64 @@
+export const searchLogURL = '/api/v1/log/entries/retrieve';
```
bdehamer, line 1: I think I prefer all of this test data to live in the test file itself.

bdehamer (changes requested):
A couple of things to fix-up described below. Also, let's not introduce that "tests" directory under the "client" directory. Ultimately, I think we'll probably move all the tests under a top-level test directory that mirrors the structure of the "src" dir, but for now let's just leave the tests alongside the code under test.

bdehamer (approved):
Nice! :shipit:

PR #64, opened 2022-08-25
proper signing/verification for attestations

bdehamer:
Closes https://github.com/sigstore/sigstore-js/issues/65

#### Summary

Previously, the only type of Rekor entry which could be created was "hashedrekord". This change provides a new signing path which creates an "intoto" Rekor entry and returns a DSSE-style sigstore bundle. A corresponding verify function has been added which accepts a DSSE bundle.

At some point we'll want to add support for all of the various Rekor entry kinds, but the "hashedrekord" and "intoto" ones are the most important for now (in support of blob and intoto provenance signing).

#### Release Note

* `Signer` class updated with a new `signAttestation` method which writes an "intoto" entry and returns a DSSE sigstore bundle.
* `Signer` class updated to rename the `sign` method to `signBlob`. The return type was renamed to `SigstoreBlobBundle` to distinguish it from the `SigstoreDSSEBundle`.
* `Verifier` class updated with a new `verifyDSSE` which performs validation on a DSSE sigstore bundle (note that the only verification done at the moment is to check the signature against the signing certificate -- lots of work to do here still).
* The "sigstore" module (the primary, public interface for this package) was updated to include new `signAttestation` and `verifyDSSE` functions.
* All types and helpers related to the sigstore signature bundles were extracted into the "bundle" module (the [bundle format](https://docs.google.com/document/d/1gucjOA_bGyRjK6TeaOI-X5GIUv8WsPzeMDMkq25Kv4Y/edit#heading=h.700g68d1uga) is still in flux -- this will likely get refactored a bit in a future update).


src/cli/index.ts
```diff
@@ -49,7 +49,11 @@ async function sign(artifactPath: string) {
 
 async function signDSSE(artifactPath: string, payloadType: string) {
   const buffer = fs.readFileSync(artifactPath);
-  const bundle = await dsse.sign(buffer, payloadType, signOptions);
+  const bundle = await sigstore.signAttestation(
+    buffer,
+    payloadType,
```
eddiezane, line 54: Should the payload type default to `application/vnd.in-toto+json`?

https://github.com/in-toto/attestation/blob/main/spec/README.md#envelope

src/util.ts
```diff
@@ -58,3 +58,12 @@ export const getUserAgent = (): string => {
   const archName = os.arch();
   return `sigstore-js/${packageVersion} (Node ${nodeVersion}) (${platformName}/${archName})`;
 };
+
+// DSSE Pre-Authentication Encoding
+export function pae(payloadType: string, payload: Buffer): Buffer {
```
eddiezane, line 63: I tried to find another home for this and understand why you landed it here. I wish there was a better place for it to live. Or maybe a name that's more clear?

eddiezane (commented):
This looks great! A few questions.

src/bundle.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export type SigstoreBundle = SigstoreBlobBundle | SigstoreDSSEBundle;
+
+export const buildBlobBundle = (
+  digest: string,
+  signature: string,
+  certificate: string,
+  rekorEntry: Entry
+): SigstoreBlobBundle => ({
```
eddiezane, line 62: cc @patflynn @znewman01 implementation here for y'all.

src/verify.ts
```diff
@@ -36,7 +38,18 @@ export class Verifier {
...
-    return verifyBlob(certificate, payload, signature);
+    return crypto.verifyBlob(certificate, payload, signature);
+  }
+
+  public async verifyDSSE(bundle: SigstoreDSSEBundle): Promise<boolean> {
+    const payloadType = bundle.attestation.payloadType;
+    const payload = Buffer.from(bundle.attestation.payload, 'base64');
+    const paeBuffer = pae(payloadType, payload);
+
+    const signature = bundle.attestation.signatures[0].sig;
```
eddiezane, line 49: Potential for an oob here. Does it matter? The error is being swallowed at https://github.com/sigstore/sigstore-js/pull/64/files#diff-2304ffbbb1cb4987906a47b08b7c3b590fc542f4247b733844049776fd8591f2R79-R84

eddiezane: Or this doesn't matter in JS...

src/cli/index.ts
bdehamer: Yes, great idea!

src/util.ts
bdehamer: Yeah, sorta hate the "util" junk drawer, but it's the best I've got for now. The bad name is a good call out though. Updated to something more descriptive.

src/verify.ts
bdehamer: Fixed

eddiezane (approved):
Awesome!

src/bundle.ts
patflynn: thanks for the pointer! Did you all take a look at https://github.com/sigstore/cosign/pull/2204 ?

bdehamer: @patflynn yeah, already working on a follow-up to bring this in line with the latest bundle work 👍

patflynn:
Awesome work!

PR #67, opened 2022-08-26
set-up codegen from sigstore bundle protobufs

bdehamer:

#### Summary

Adds some plumbing to automatically codegen types for the sigstore bundle format from the published protobufs.

Types can be re-generated by executing the following:
```
npm run codegen:bundle
```

The generated types are written to `src/bundle/__generated__`.

#### Release Note

* New script in `package.json` for re-generating bundle types from their protobufs


bdehamer:
cc @kommendorkapten @patflynn generating TypeScript types from the protobufs in https://github.com/sigstore/cosign/pull/2204

bin/generate-bundle-types
```diff
@@ -0,0 +1,29 @@
...
+
+curl -L --silent --show-error \
+  -o /tmp/defs/envelope.proto \
+  https://raw.githubusercontent.com/${REPO}/${REF}/${PB_PATH}/envelope.proto
+
+curl -L --silent --show-error \
+  -o /tmp/defs/bundle.proto \
+  https://raw.githubusercontent.com/${REPO}/${REF}/${PB_PATH}/bundle.proto
+
+protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto \
```
feelepxyz, line 18: Does this need to be installed separately? Should we set up a Dockerfile for this?

feelepxyz: Realised running this in docker is not entirely trivial.. can have a crack at this if helpful as I kind of started poking at this while watching MS training videos 😄  would this be helpful?

bdehamer: @feelepxyz Yeah, if you wanna take a crack at this, that would be great. I wasn't sure that if asking the dev to have docker installed was any less burdensome than just installing `protoc` 😆

feelepxyz: > I wasn't sure that if asking the dev to have docker installed was any less burdensome than just installing `protoc` 

This is a good point, I think docker is probably fairly common though. I noticed that using an older version of protoc generated slightly different output which would not be the case if we pin the version in the docker image.

Will also add some docs/readme on installing docker as a prerequisite for generating the bundle protobufs.

bin/generate-bundle-types
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
```
feelepxyz, line 1: I'm not 100% sold on the pattern here as it feels a bit surprising but went with something we did on a couple Dependabot projects. Here all you need to do is have docker installed and run `bin/generate-bundle-types` and should "just work".

I moved the curl downloads out from and set them up in the `Dockerfile` so we can run this bin script without network access 🔒

Dockerfile.bundle-types
```diff
@@ -0,0 +1,42 @@
...
+  && rm /tmp/protoc.zip
+
+###############
+FROM node:16.17
+
+WORKDIR /client
+
+# Download proto definitions for .sigstore bundle from sigstore/cosign
+ARG REPO=sigstore/cosign
+ARG REF=a8ea54021bb6f850b30ee22bd15e0477d623e00d
```
feelepxyz, line 24: Bleeding edge 😄  looks like this is referencing https://github.com/sigstore/cosign/pull/2204/files

README.md
```diff
@@ -22,6 +22,33 @@ A tool for signing and verifying npm packages.
 * Support offline and online signature verification
 * Retrieve root cert and public keys via TUF client
 
+## Prerequisites
+
+- Node.js version 16+ (LTS)
+- Add `node_modules/.bin` to your `$PATH` to access the installed `tsc` bin
```
feelepxyz, line 28: @bdehamer added these basic setup requirements while adding the docker prereq. Is this what you ended up doing as well with `PATH`?

feelepxyz:
@bdehamer just force pushed as I forgot to add sig-off info on the commits.

feelepxyz:
@eddiezane 👋 gentle bump, did you have any ideas for improvements here?

README.md
bdehamer: I've never done this. I think the convention here is to just use `npx` (`npm exec`) to invoke project-local binaries.

hack/images/protoc/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+###############
+FROM node:16.17
+
+WORKDIR /client
+
+COPY --from=protoc /protoc/bin/protoc /usr/local/bin
+
+ARG TSPROTO_VERSION=1.125.0
+RUN npm install -g "ts-proto@${TSPROTO_VERSION}"
```
feelepxyz, line 25: Installing this explicitly is much nicer 👍

feelepxyz (approved):
@bdehamer ace refactor of my codes 👍 💯  much easier to follow.

PR #68, opened 2022-08-29
Re-add logging of Rekor entry index and url

danbev:
#### Summary
Currently, the README.md file refers to output from the sign command:
```console
Created entry at index 2698234, available at
https://rekor.sigstore.dev/api/v1/log/entries/43553c769...
```
It looks like this output was removed in Commit bd596c0eccd6104589e596c34936a6c46a2e30ae ("proper signing/verification for attestations").

This commit suggest re-adding this output as I was not sure how to get
to this information otherwise. I'm very new to this code base but
hopefully someone can provide some feedback if there are other ways,
or if the README.md should be updated instead.

#### Release Note
N/A

#### Documentation
N/A

Signed-off-by: Daniel Bevenius <daniel.bevenius@gmail.com>

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -47,6 +47,10 @@ async function sign(artifactPath: string) {
   const buffer = fs.readFileSync(artifactPath);
   const bundle = await sigstore.sign(buffer, signOptions);
   console.log(JSON.stringify(bundle));
+
+  const url = `${sigstore.getRekorBaseUrl(signOptions)}/api/v1/log/entries`;
+  console.error(`Created entry at index ${bundle.logIndex}, available at`);
+  console.error(`${url}?logIndex=${bundle.logIndex}`);
 }
 
 async function signDSSE(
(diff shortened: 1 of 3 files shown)


bdehamer (changes requested):
@danbev Thanks for the contribution! I had only removed that output cause I didn't want to be writing to the console from within the Rekor client like that. However, it is definitely useful to have that info available when you're invoking this from the CLI . . . so let's move this into the `cli` module. 

I don't think we have the log entry UUID available at that point, but it's still possible to construct a Rekor URL using the `logIndex`:

```
https://rekor.sigstore.dev/api/v1/log/entries?logIndex=${logIndex}
```

danbev:
> However, it is definitely useful to have that info available when you're invoking this from the CLI . . . so let's move this into the `cli` module.

Ah that makes sense. I'll update this shortly. Thanks!

bdehamer (changes requested):
Looks great! 

Only request is to remove the `src/cli/.index.ts.swo` file (guessing this was committed by mistake)

danbev:
> Only request is to remove the `src/cli/.index.ts.swo` file (guessing this was committed by mistake)

Oh sorry about that, yes that was a mistake. I'll remove it.

bdehamer (approved):
👍

PR #84, opened 2022-09-16
Split the cert chain returned from Fulcio

bdehamer:

#### Summary
The certificate we receive from Fulcio is actually a chain of certificates including the signing certificate, an intermediate certificate, and the Fulcio root certificate. We've been sending all of these to Rekor when we really only need the leaf signing certificate.

This adds some code to split the certificate chain and send only the signing certificate to Rekor.

#### Release Note
* When creating an entry in Rekor, sends only the leaf signing certificate instead of the entire Fulcio chain.


src/sign.ts
```diff
@@ -56,7 +57,7 @@ export class Signer {
     // Calculate artifact digest
     const digest = crypto.hash(payload);
 
-    const certificateB64 = enc.base64Encode(certificate);
+    const certificateB64 = enc.base64Encode(certificate[0]);
```
feelepxyz, line 60: nit could we make it more explicit we are fetching the leaf here? e.g.

```suggestion
    const = leafCertificate = certificate[0]
    const certificateB64 = enc.base64Encode(leafCertificate);
```

src/sign.ts
```diff
@@ -89,7 +90,7 @@ export class Signer {
       ],
     };
 
-    const certificateB64 = enc.base64Encode(certificate);
+    const certificateB64 = enc.base64Encode(certificate[0]);
```
feelepxyz, line 93:
```suggestion
    const = leafCertificate = certificate[0]
    const certificateB64 = enc.base64Encode(leafCertificate);
```

feelepxyz (dismissed):
👍  LGTM. Added  a comment about making the variables a bit more explicit. Might be easy to miss that we're pulling out the first one.

src/sign.ts
```diff
@@ -20,11 +20,12 @@ import {
   SigstoreBlobBundle,
```
feelepxyz, line 20: Would be good to add a test case to the signBlob and signAttestation expecting the leaf cert to be sent to rekor

bdehamer: Good idea. I updated one of the tests to check this.

src/sign.ts
```diff
@@ -89,7 +91,7 @@ export class Signer {
       ],
     };
 
-    const certificateB64 = enc.base64Encode(certificate);
+    const certificateB64 = enc.base64Encode(certificate[0]);
```
feelepxyz, line 94: Nit should this also be 
```suggestion
    const leafCertificate = certificate[0];
    const leafCertificateB64 = enc.base64Encode(leafCertificate);
```

feelepxyz (approved):

PR #107, opened 2022-10-11
Update sign/verify to take the bundle as json

feelepxyz:
#### Summary

Updated the `sign/signAttestation/verify` library methods to take and return the bundle file as json to match the format on disk.

With the current interface a caller would be forced to convert the json to the Bundle object format which seems like an implementation detail. I ran into this implementing bundle verification in npm where we want to use the `Sigstore.verify` method to verify the bundle.

Opening up for comments/feedback.

#### Release Note

* **BREAKING CHANGES**
  * `sigstore.sign(...): Promise<Bundle> > sigstore.sign(...): Promise<object>`: Promise returns a json object where it previously returned a Bundle object
  * `sigstore.signAttestation(...): Promise<Bundle> > sigstore.signAttestation(...): Promise<object>`: Promise returns a json object where it previously returned a Bundle object
  * `sigstore.verify(bundle: Bundle, data?: Buffer, options: {}) > sigstore.verify(bundle: object, data?: Buffer, options: {})`: Verify now takes a the bundle as json object instead of a Bundle object

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,9 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:ci": "jest --maxWorkers=2 --coverage",
-    "lint": "eslint --fix --ext .ts src/**",
-    "lint:check": "eslint --max-warnings 0 --ext .ts src/**",
-    "format": "prettier --write \"src/**/*\"",
+    "lint": "eslint --fix --ignore-path .lintignore --ext .ts src",
+    "lint:check": "eslint --ignore-path .lintignore --max-warnings 0 --ext .ts src",
+    "format": "prettier --write --ignore-path .lintignore \"src\"",
     "codegen:bundle": "./hack/generate-bundle-types",
     "codegen:rekor": "./hack/generate-rekor-types"
   },
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import fs from 'fs';
-import { Bundle } from '../types/bundle';
 import { sigstore } from '../index';
 
 const INTOTO_PAYLOAD_TYPE = 'application/vnd.in-toto+json';
@@ -42,15 +41,14 @@ const signOptions = {
 
 async function sign(artifactPath: string) {
   const buffer = fs.readFileSync(artifactPath);
-  const bundleJson = await sigstore.sign(buffer, signOptions);
+  const bundle: any = await sigstore.sign(buffer, signOptions); //eslint-disable-line @typescript-eslint/no-explicit-any
 
   const url = `${sigstore.getRekorBaseUrl(signOptions)}/api/v1/log/entries`;
-  const bundle = Bundle.fromJSON(JSON.parse(bundleJson));
   const logIndex = bundle.timestampVerificationData?.tlogEntries[0].logIndex;
   console.error(`Created entry at index ${logIndex}, available at`);
   console.error(`${url}?logIndex=${logIndex}`);
 
-  console.log(bundleJson);
+  console.log(JSON.stringify(bundle));
 }
 
 async function signDSSE(
@@ -74,7 +72,10 @@ async function verify(bundlePath: string, artifactPath: string) {
   }
 
   const bundleFile = fs.readFileSync(bundlePath);
-  const result = await sigstore.verify(bundleFile.toString('utf-8'), payload);
+  const result = await sigstore.verify(
+    JSON.parse(bundleFile.toString('utf-8')),
+    payload
+  );
 
   if (result) {
     console.error('Verified OK');
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -42,14 +42,15 @@ const signOptions = {
 
 async function sign(artifactPath: string) {
   const buffer = fs.readFileSync(artifactPath);
-  const bundle = await sigstore.sign(buffer, signOptions);
+  const bundleJson = await sigstore.sign(buffer, signOptions);
 
   const url = `${sigstore.getRekorBaseUrl(signOptions)}/api/v1/log/entries`;
+  const bundle = Bundle.fromJSON(JSON.parse(bundleJson));
   const logIndex = bundle.timestampVerificationData?.tlogEntries[0].logIndex;
   console.error(`Created entry at index ${logIndex}, available at`);
   console.error(`${url}?logIndex=${logIndex}`);
 
-  console.log(JSON.stringify(Bundle.toJSON(bundle)));
+  console.log(bundleJson);
 }
 
 async function signDSSE(
@@ -57,12 +58,12 @@ async function signDSSE(
   payloadType = INTOTO_PAYLOAD_TYPE
 ) {
   const buffer = fs.readFileSync(artifactPath);
-  const bundle = await sigstore.signAttestation(
+  const bundleJson = await sigstore.signAttestation(
     buffer,
     payloadType,
     signOptions
   );
-  console.log(JSON.stringify(Bundle.toJSON(bundle)));
+  console.log(bundleJson);
 }
 
 async function verify(bundlePath: string, artifactPath: string) {
@@ -73,9 +74,7 @@ async function verify(bundlePath: string, artifactPath: string) {
   }
 
   const bundleFile = fs.readFileSync(bundlePath);
-  const bundle = Bundle.fromJSON(JSON.parse(bundleFile.toString('utf-8')));
-
-  const result = await sigstore.verify(bundle, payload);
+  const result = await sigstore.verify(bundleFile.toString('utf-8'), payload);
 
   if (result) {
     console.error('Verified OK');
--- a/src/sigstore.test.ts
+++ b/src/sigstore.test.ts
@@ -24,9 +24,30 @@ describe('sign', () => {
   const payload = Buffer.from('Hello, world!');
 
   // Signer output
-  const signedPayload = {
-    base64Signature: 'signature',
-    cert: 'cert',
+  const signedPayload: Bundle = {
+    mediaType: 'application/vnd.dev.sigstore.bundle+json;version=0.1',
+    content: {
+      $case: 'messageSignature',
+      messageSignature: {
+        messageDigest: {
+          algorithm: HashAlgorithm.SHA2_256,
+          digest: Buffer.from(''),
+        },
+        signature: Buffer.from(''),
+      },
+    },
+    timestampVerificationData: {
+      rfc3161Timestamps: [],
+      tlogEntries: [],
+    },
+    verificationMaterial: {
+      content: {
+        $case: 'x509CertificateChain',
+        x509CertificateChain: {
+          certificates: [],
+        },
+      },
+    },
   };
 
   const mockSigner = jest.mocked(Signer);
@@ -65,8 +86,7 @@ describe('sign', () => {
 
   it('returns the correct envelope', async () => {
     const sig = await sign(payload);
-
-    expect(sig).toEqual(signedPayload);
+    expect(sig).toMatchSnapshot();
   });
 });
 
@@ -76,8 +96,32 @@ describe('signAttestation', () => {
 
   // Signer output
   const signedPayload = {
-    base64Signature: 'signature',
-    cert: 'cert',
+    mediaType: 'application/vnd.dev.sigstore.bundle+json;version=0.1',
+    content: {
+      $case: 'dsseEnvelope',
+      dsseEnvelope: {
+        payload: Buffer.from(''),
+        payloadType: 'text/json',
+        signatures: [
+          {
+            sig: Buffer.from(''),
+            keyid: 'key-123',
+          },
+        ],
+      },
+    },
+    timestampVerificationData: {
+      rfc3161Timestamps: [],
+      tlogEntries: [],
+    },
+    verificationMaterial: {
+      content: {
+        $case: 'x509CertificateChain',
+        x509CertificateChain: {
+          certificates: [],
+        },
+      },
+    },
   };
 
   const mockSigner = jest.mocked(Signer);
@@ -118,8 +162,7 @@ describe('signAttestation', () => {
 
   it('returns the correct envelope', async () => {
     const sig = await signAttestation(payload, payloadType);
-
-    expect(sig).toEqual(signedPayload);
+    expect(sig).toMatchSnapshot();
   });
 });
 
@@ -149,6 +192,7 @@ describe('#verify', () => {
       },
     },
   };
+  const bundleJson = JSON.stringify(Bundle.toJSON(bundle));
 
   const mockVerify = jest.fn();
 
@@ -159,12 +203,12 @@ describe('#verify', () => {
   });
 
   it('invokes the Verifier instance with the correct params', async () => {
-    await verify(bundle);
+    await verify(bundleJson);
     expect(mockVerify).toHaveBeenCalledWith(bundle, undefined);
   });
 
   it('returns the value returned by the verifier', async () => {
-    const result = await verify(bundle);
+    const result = await verify(bundleJson);
     expect(result).toBe(false);
   });
 });
--- a/src/sigstore.test.ts
+++ b/src/sigstore.test.ts
@@ -86,7 +86,7 @@ describe('sign', () => {
 
   it('returns the correct envelope', async () => {
     const sig = await sign(payload);
-    expect(sig).toMatchSnapshot();
+    expect(JSON.stringify(sig)).toMatchSnapshot();
   });
 });
 
@@ -162,7 +162,7 @@ describe('signAttestation', () => {
 
   it('returns the correct envelope', async () => {
     const sig = await signAttestation(payload, payloadType);
-    expect(sig).toMatchSnapshot();
+    expect(JSON.stringify(sig)).toMatchSnapshot();
   });
 });
 
@@ -192,7 +192,6 @@ describe('#verify', () => {
       },
     },
   };
-  const bundleJson = JSON.stringify(Bundle.toJSON(bundle));
 
   const mockVerify = jest.fn();
 
@@ -203,12 +202,13 @@ describe('#verify', () => {
   });
 
   it('invokes the Verifier instance with the correct params', async () => {
-    await verify(bundleJson);
+    const b = Bundle.toJSON(bundle);
+    await verify(b);
     expect(mockVerify).toHaveBeenCalledWith(bundle, undefined);
   });
 
   it('returns the value returned by the verifier', async () => {
-    const result = await verify(bundleJson);
+    const result = await verify(Bundle.toJSON(bundle));
     expect(result).toBe(false);
   });
 });
--- a/src/sigstore.ts
+++ b/src/sigstore.ts
@@ -44,41 +44,49 @@ type IdentityProviderOptions = Pick<
 export async function sign(
   payload: Buffer,
   options: SignOptions = {}
-): Promise<Bundle> {
+): Promise<string> {
   const fulcio = new Fulcio({ baseURL: options.fulcioBaseURL });
   const rekor = new Rekor({ baseURL: options.rekorBaseURL });
   const idps = configureIdentityProviders(options);
+  console.log('Signed payload: ', payload);
 
-  return new Signer({
-    fulcio,
-    rekor,
-    identityProviders: idps,
-  }).signBlob(payload);
+  return new Signer({ fulcio, rekor, identityProviders: idps })
+    .signBlob(payload)
+    .then((bundle: Bundle) => {
+      console.log('Signed bundle: ', bundle);
+
+      return JSON.stringify(Bundle.toJSON(bundle))
+    });
 }
 
 export async function signAttestation(
   payload: Buffer,
   payloadType: string,
   options: SignOptions = {}
-): Promise<Bundle> {
+): Promise<string> {
   const fulcio = new Fulcio({ baseURL: options.fulcioBaseURL });
   const rekor = new Rekor({ baseURL: options.rekorBaseURL });
   const idps = configureIdentityProviders(options);
 
-  return new Signer({
-    fulcio,
-    rekor,
-    identityProviders: idps,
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -9,9 +9,9 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:ci": "jest --maxWorkers=2 --coverage",
-    "lint": "eslint --fix --ext .ts src/**",
-    "lint:check": "eslint --max-warnings 0 --ext .ts src/**",
-    "format": "prettier --write \"src/**/*\"",
+    "lint": "eslint --fix --ignore-path .lintignore --ext .ts src",
+    "lint:check": "eslint --ignore-path .lintignore --max-warnings 0 --ext .ts src",
+    "format": "prettier --write --ignore-path .lintignore \"src\"",
```
feelepxyz, line 14: `src/**` was causing these tools to try and find `.ts` files in the new src/__snapshots__ folder and failing.

src/sigstore.test.ts
```diff
@@ -65,8 +86,7 @@ describe('sign', () => {
 
   it('returns the correct envelope', async () => {
     const sig = await sign(payload);
-
-    expect(sig).toEqual(signedPayload);
+    expect(sig).toMatchSnapshot();
```
feelepxyz, line 89: Don't love these kinds of tests as I think it's easy for breaking changes to sneak past undetected but also painful to write out the entire json string here.

src/sigstore.ts
```diff
@@ -44,41 +44,41 @@ type IdentityProviderOptions = Pick<
 export async function sign(
   payload: Buffer,
   options: SignOptions = {}
-): Promise<Bundle> {
+): Promise<string> {
```
feelepxyz, line 47: I was going back and forward on this change, initially I thought that maybe we should expose a bundle serialiser/deserialiser method on the `sigstore` export but exposing the `Bundle` interface didn't feel quite right as the format includes implementation details from the proto <> typescript conversion.

But then again, if you used this lib from typescript it would be nice to get a typed object back! Thoughts on this?

feelepxyz: also cc @kommendorkapten as sigstore bundle implementor

bdehamer (commented):
This doesn't feel quite right to me. I think that these interfaces should treat the `Bundle` as the primary input/output. I like the consistency of both `sign` and `verify` trafficking in the same data types and having `sign` return a string seems unfriendly to the consuming code. 

The change made to the CLI which immediately re-hydrates the `Bundle` immediately after it was serialized sorta illustrates the issue w/ this approach.

Can we split the difference here and support both cases? The `verify` function could easily support `string | Bundle` and the `sign` functions could probably use an input option to indicate the preferred output format.

src/sigstore.ts
kommendorkapten: While I'm not a JS/TS expert, I do feel that the sigstore-js library should acept/produce bundle objects, as that's what's native to the APIs? I think internally sigstore-js should not perform any operations on transforming data more than really necessarily, if client's prefer to use a diferent object model, I think then clients can implement that wrapper layer as that would be a domain logic decision?

feelepxyz: > sigstore-js library should acept/produce bundle objects

You mean the json representation on disk or a typed bundle object? Currently the API requires the caller to transform the json representation into a `Bundle` interface by calling the `Bundle.fromJSON` method, which is internal to the library and generated from protobufs in cosign.

feelepxyz:
> I think that these interfaces should treat the `Bundle` as the primary input/output

I'm not sure about this for the external interface. If you use the lib from a typescript codebase you would be mostly good as the type system would tell which attributes exist on the returned bundle. The problem would still be that you can't just `JSON.stringify` the returned bundle and persist it as a valid `.sigstore` bundle so would need to export a serializer/deserializer that calls the `Bundle.to/fromJSON` method. I think the disconnect between the `.sigstore` bundle format and the typescript interface would only cause confusion when using this as a library.

Should we not treat the `Bundle` interface as internal to the library and maybe change this to take a JS object instead of a string so the caller is responsible for `JSON.parse/stringify`?

It would be neat to have a typescript interface that matches the json object in the `.sigstore` bundle but not sure we can autogenerate this?

src/cli/index.ts
```diff
@@ -42,27 +41,27 @@ const signOptions = {
 
 async function sign(artifactPath: string) {
   const buffer = fs.readFileSync(artifactPath);
-  const bundle = await sigstore.sign(buffer, signOptions);
+  const bundle: any = await sigstore.sign(buffer, signOptions); //eslint-disable-line @typescript-eslint/no-explicit-any
```
feelepxyz, line 44: 😞

src/sigstore.ts
kommendorkapten: For JS/TS I would have to let @bdehamer answer what's the common naming schema when using protoserializers. But  I'm not sure that sigstore-js itself should expose those methods? I would think that marshalling is not something that sigstore-js is responsible for, but a separate package exists, that defines the data objects, and any serialization methods. 

The approach I (and some others) are to create a new repo in sigstore, `sigstore/protobuf` where the actual protobuf files exist, and also it's from there where all the language clients are built, and published to the appropriate repository. Then sigstore-js can import the package from there.

feelepxyz: > Then sigstore-js can import the package from there.

Hm this could be a good longer term aim but we'll need to figure something out to be able to use this library from npm land. Currently blocked on this.

@bdehamer thoughts on exposing some methods to serialize/deserialize the bundle so we can use it with the json bundle file?

bdehamer:
> This typed bundle <> json disconnect is clearly still breaking my brain in JS where JSON is a serialized javascript object

Yeah, I totally agree -- I've been wrestling with this same thing. Our chosen serialization format (JSON) **is the native data format** in JavaScript, so why wouldn't we just use that as the internal Bundle representation, right? The protobuf and the associated code generation are nice convenience, but ultimately result in a data structure that feels a bit foreign in the JS world . . . and requires some awkward marshalling between **different** JSON representations.

I see 3 potential paths forward:

1. Move ahead w/ the protobuf code-generated `Bundle` object as the public interface. Export the necessary `toJSON`/`fromJSON` helpers and instruct consumers to serialize/deserialize as-needed.
2. Ditch the code-generation altogether. Hand-craft a new `Bundle` interface that matches the serialized form of the protobuf. This gives us an in-memory representation that perfectly matches the serialized structure, but we lose the benefits of code-generation . . . and potentially creates a bunch more work for us in the future if/when Fulcio/Rekor have GRPC APIs which traffic in the protobuf messages we've abandoned.
3. Use the code-generated interface internally, but change the public interface to use the serialization-JSON format instead of the protobuf-JSON format (this is essentially what @feelepxyz has done it the most recent commit on this PR). Instead of `any`-typing this though I'd want to define a true Typescript interface which describes the shape of this data . . . but this requires that we hand-craft an interface that we now need to keep in-sync w/ the code-generated one.

Honestly, #<span/>2 probably doesn't make sense . . . if we're going to commit to maintaining a hand-crafted interface which matches the serialization-JSON format, we might as well do #<span/>3 and keep some of the benefits of the codegen.

bdehamer:
Adding some more context for anyone who hasn't interacted with this code directly and may not grok the issue @feelepxyz is raising.

As a result of code-generating our data structures from the Sigstore Bundle protobufs we end-up with two different JSON representations of a bundle that we're juggling: The protobuf-JSON which is code-generated directly from the protocol buffers, and the serialization-JSON which is the result of applying the standard protobuf JSON serialization rules to the protobuf-JSON form.

**protobuf-JSON**
```js
{
  mediaType: 'application/vnd.dev.sigstore.bundle+json;version=0.1',
  content: {
    '$case': 'messageSignature',
    messageSignature: {
      messageDigest: {
        algorithm: 1,
        digest: <Buffer 37 02 6c ca 0d e0 08 0a 8d aa 67 ea 24 96 21 8b 69 22 66 03 66 25 04 1c 38 25 45 57 ff d9 9e 81>
      },
      signature: <Buffer 30 46 02 21 00 d8 3e ab 0b d1 62 1d bb 7f 19 9e ef 28 ad 30 8b 20 ec 1a e2 af 65 fb e7 f5 67 ff 8c e5 59 f6 46 02 21 00 a8 77 94 2e 30 89 34 79 5d 3a ... 22 more bytes>
    }
  },
  timestampVerificationData: { 
    tlogEntries: [ 
      {
        inclusionPromise: <Buffer 30 45 02 20 4e 94 d9 94 97 7a 0f 49 0b 8d f2 a3 4f 82 ab 68 fd df 7e d8 04 55 5a 65 26 f6 6e e3 6e 85 87 ab 02 21 00 af 31 8b fb 19 f3 68 7b 7c 45 22 ... 21 more bytes>,
        logIndex: '5041381',
        logId: <Buffer c0 d2 3d 6a d4 06 97 3f 95 59 f3 ba 2d 1c a0 1f 84 14 7d 8f fc 5b 84 45 c2 24 f9 8b 95 91 80 1d>,
        integratedTime: '1665683100',
        kindVersion: { kind: 'hashedrekord', version: '0.0.1' },
        inclusionProof: undefined
      }
    ], 
    rfc3161Timestamps: []
  },
  verificationMaterial: {
    content: { 
      '$case': 'x509CertificateChain', 
      x509CertificateChain: {
        certificates: [
          <Buffer 30 82 02 a0 30 82 02 26 a0 03 02 01 02 02 14 56 6b 42 56 e1 b0 7a e5 f5 b2 d5 20 df 38 b0 67 27 94 d1 b7 30 0a 06 08 2a 86 48 ce 3d 04 03 03 30 37 31 ... 626 more bytes>,
          <Buffer 30 82 02 1a 30 82 01 a1 a0 03 02 01 02 02 14 00 b9 d5 89 57 e7 53 46 eb 25 ab 26 46 41 eb 9f f5 27 7d a4 30 0a 06 08 2a 86 48 ce 3d 04 03 03 30 2a 31 ... 492 more bytes>,
          <Buffer 30 82 01 f7 30 82 01 7c a0 03 02 01 02 02 14 00 b6 4d 00 f1 5d c4 73 f0 8d e0 e5 a0 3c 32 60 28 40 3b fe 30 0a 06 08 2a 86 48 ce 3d 04 03 03 30 2a 31 ... 457 more bytes>
        ]
      }
    }
  }
}
```

**serialization-JSON**
```json
{
  "mediaType": "application/vnd.dev.sigstore.bundle+json;version=0.1",
  "timestampVerificationData": {
    "tlogEntries": [
      {
        "logIndex": "5041381",
        "logId": "wNI9atQGlz+VWfO6LRygH4QUfY/8W4RFwiT5i5WRgB0=",
        "kindVersion": {
          "kind": "hashedrekord",
          "version": "0.0.1"
        },
        "integratedTime": "1665683100",
        "inclusionPromise": "MEUCIE6U2ZSXeg9JC43yo0+Cq2j9337YBFVaZSb2buNuhYerAiEArzGL+xnzaHt8RSJ5b4sZs4Mcsz29f+OeqDLVJCQgbyc="
      }
    ],
    "rfc3161Timestamps": []
  },
  "verificationMaterial": {
    "x509CertificateChain": {
      "certificates": [
        "MIICoDCCAiagAwIBAgIUVmtCVuGweuX1stUg3ziwZyeU0bcwCgYIKoZIzj0EAwMwNzEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MR4wHAYDVQQDExVzaWdzdG9yZS1pbnRlcm1lZGlhdGUwHhcNMjIxMDEzMTc0NDU5WhcNMjIxMDEzMTc1NDU5WjAAMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEzGwWp5Mfn1SVMhIbOcyoXIPel0/34oboTLYaDe5QkxidoKBG9SQ5IV9df9HzWPYebQDlm4O1XcuMTCXDKJaN/KOCAUUwggFBMA4GA1UdDwEB/wQEAwIHgDATBgNVHSUEDDAKBggrBgEFBQcDAzAdBgNVHQ4EFgQUaTDZ6d8Oqi1gSVgeXCB0LlZZvskwHwYDVR0jBBgwFoAU39Ppz1YkEZb5qNjpKFWixi4YZD8wHwYDVR0RAQH/BBUwE4ERYnJpYW5AZGVoYW1lci5jb20wLAYKKwYBBAGDvzABAQQeaHR0cHM6Ly9naXRodWIuY29tL2xvZ2luL29hdXRoMIGKBgorBgEEAdZ5AgQCBHwEegB4AHYACGCS8ChS/2hF0dFrJ4ScRWcYrBY9wzjSbea8IgY2b3IAAAGD0nMP/gAABAMARzBFAiEA2tevNvoP7S3dEY1NVB0waPdu9PQkNiN8U+32U2qU13MCIAy6hatFLWzRLrIb84Vtm5Q8T1Y7x7ArngISi7dtbjjlMAoGCCqGSM49BAMDA2gAMGUCMQCcm6DpsymeUPzNhdKE9FbiOfuFqp4BFPdMzVju3eJAUq+5/AgilZv+8FwI7sEOXa0CMCsP17jfLRvxP2HKOlj6qfCBBqKujTATbmA6EuhvWlQ4nV7UhzAvPyICdsyJDrm2HQ==",
        "MIICGjCCAaGgAwIBAgIUALnViVfnU0brJasmRkHrn/UnfaQwCgYIKoZIzj0EAwMwKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0yMjA0MTMyMDA2MTVaFw0zMTEwMDUxMzU2NThaMDcxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjEeMBwGA1UEAxMVc2lnc3RvcmUtaW50ZXJtZWRpYXRlMHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE8RVS/ysH+NOvuDZyPIZtilgUF9NlarYpAd9HP1vBBH1U5CV77LSS7s0ZiH4nE7Hv7ptS6LvvR/STk798LVgMzLlJ4HeIfF3tHSaexLcYpSASr1kS0N/RgBJz/9jWCiXno3sweTAOBgNVHQ8BAf8EBAMCAQYwEwYDVR0lBAwwCgYIKwYBBQUHAwMwEgYDVR0TAQH/BAgwBgEB/wIBADAdBgNVHQ4EFgQU39Ppz1YkEZb5qNjpKFWixi4YZD8wHwYDVR0jBBgwFoAUWMAeX5FFpWapesyQoZMi0CrFxfowCgYIKoZIzj0EAwMDZwAwZAIwPCsQK4DYiZYDPIaDi5HFKnfxXx6ASSVmERfsynYBiX2X6SJRnZU84/9DZdnFvvxmAjBOt6QpBlc4J/0DxvkTCqpclvziL6BCCPnjdlIB3Pu3BxsPmygUY7Ii2zbdCdliiow=",
        "MIIB9zCCAXygAwIBAgIUALZNAPFdxHPwjeDloDwyYChAO/4wCgYIKoZIzj0EAwMwKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0yMTEwMDcxMzU2NTlaFw0zMTEwMDUxMzU2NThaMCoxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjERMA8GA1UEAxMIc2lnc3RvcmUwdjAQBgcqhkjOPQIBBgUrgQQAIgNiAAT7XeFT4rb3PQGwS4IajtLk3/OlnpgangaBclYpsYBr5i+4ynB07ceb3LP0OIOZdxexX69c5iVuyJRQ+Hz05yi+UF3uBWAlHpiS5sh0+H2GHE7SXrk1EC5m1Tr19L9gg92jYzBhMA4GA1UdDwEB/wQEAwIBBjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBRYwB5fkUWlZql6zJChkyLQKsXF+jAfBgNVHSMEGDAWgBRYwB5fkUWlZql6zJChkyLQKsXF+jAKBggqhkjOPQQDAwNpADBmAjEAj1nHeXZp+13NWBNa+EDsDP8G1WWg1tCMWP/WHPqpaVo0jhsweNFZgSs0eE7wYI4qAjEA2WB9ot98sIkoF3vZYdd3/VtWB5b9TNMea7Ix/stJ5TfcLLeABLE4BNJOsQ4vnBHJ"
      ]
    }
  },
  "messageSignature": {
    "messageDigest": {
      "algorithm": "SHA2_256",
      "digest": "NwJsyg3gCAqNqmfqJJYhi2kiZgNmJQQcOCVFV//ZnoE="
    },
    "signature": "MEYCIQDYPqsL0WIdu38Znu8orTCLIOwa4q9l++f1Z/+M5Vn2RgIhAKh3lC4wiTR5XToWYOixaq+8K1roQ9JF6Cd/l3zhl+cV"
  }
}
```

The question at hand is essentially: which of these two forms should be used in the public interface of the `sigstore-js` library?

src/cli/index.ts
bdehamer: Yeah, don't like this at all. If we're gonna go this route, I'd want to define a custom interface which matches the serialized form of the Bundle (see my comments below).

feelepxyz:
@bdehamer great summary of the issue with clear next steps! 💯 

> 3\. Use the code-generated interface internally, but change the public interface to use the serialization-JSON format instead of the protobuf-JSON format (this is essentially what @feelepxyz has done it the most recent commit on this PR). Instead of `any`-typing this though I'd want to define a true Typescript interface which describes the shape of this data . . . but this requires that we hand-craft an interface that we now need to keep in-sync w/ the code-generated one.

Personally I prefer this route as the current `Bundle` representation isn't protobuf, it's a conversion to typescript so we're leaking this implementation detail. I think we should either be talking bundle-JSON format or the official protobuf format, which still takes a bundle-JSON format as an input if you where to use that externally.

Forcing a consumer to install a protobuf lib seems like a no-go for the CLI so my preference would be to go with the bundle-JSON format and write a Typescript interface for it. 

If we use the code-generated interface internally we can write some tests that make sure the TS interface I/O matches the generated Bundle interface. It might still be tricky to get full coverage unless we do some fuzzying with the `Bundle` interface and auto-generate all permutations if it, convert to to JSON and test it against the TS interface.

Sidenote: if we use the bundle-JSON as input/output we could also decide to use the proto files natively with [protobuf.js](https://github.com/protobufjs/protobuf.js) internally without worrying about the interface changing. This lib uses the interface `ProtoClass.toObject/fromObject` for the serialisation. The overhead is probably not worth it for npm/cli though.

kommendorkapten:
Thanks for the clarification @bdehamer @feelepxyz, I think I understand the issue right now.

If we think the common use-case would be to JSON parse the bundle file, then use whatever object is returned to interact with sigstore-js I think it makes sense to rely on the `serialization-JSON`, and then internally convert that to `proto-JSON` (which isn't really the correct `proto-JSON` as @feelepxyz pointed out).

Would it be confusing to expose both methods? Then the implementation for `serialization-JSON` would only perform the conversion then call the method for `proto-JSON`?

kommendorkapten:
> but this requires that we hand-craft an interface that we now need to keep in-sync w/ the code-generated one.

Is this really the case? If we have the `serialization-JSON` object, which is how a proper `proto-JSON` serialization will look like, can we just make a string representation of this in memory, then call the protobuf generated `unmarshal` on that to get us a `proto-JSON` object?

feelepxyz:
> If we think the common use-case would be to JSON parse the bundle file

I think so, this is so natural in JS land so I think doing something different would cause a bunch of confusion. It's also convenient the other way around when serialising it to be sent from cli to registry from example. The caller would then just need to `POST body: JSON.stringify(sigstore.signAttestation(subject))` for example.

> then call the protobuf generated `unmarshal` on that to get us a `proto-JSON` object?

Don't think getting the `proto-JSON` is the problem. It's that the `serialization-JSON` object doesn't have a type definition so we end up with this: https://github.com/sigstore/sigstore-js/pull/107/files#diff-2304ffbbb1cb4987906a47b08b7c3b590fc542f4247b733844049776fd8591f2R44 

I would be in favour of just handcrafting a TS interface for the `serialization-JSON` object and go from there. Thoughts on this?

kommendorkapten:
> I would be in favour of just handcrafting a TS interface for the serialization-JSON object and go from there. Thoughts on this?

I don't really understand the issue. The generated code from the protobuf definition should be able to parse raw json (i.e `serialization-JSON` as documented above) into an object that has the type informaion? I'm happy to jump onto a call to discuss this more as I think I'm missing the problem.

feelepxyz:
> The generated code from the protobuf definition should be able to parse raw json (i.e `serialization-JSON` as documented above) into an object that has the type informaion?

Yep this works 👍  so internally we have a typed object and that already works. The issue is that we want to change the return value from the `sign` methods to return the raw json object/`serialization-JSON` object, but this return value no longer has any type information. We could create a TS interface that exactly matches the `serialization-JSON` representation and return this when used as a library.

Returning the `serialization-JSON` object from `sign` means that a caller can just do something like `fs.writeFileSync('bundle.sigstore', JSON.stringify(sigstore.signAttestation(subject)))` instead of something like:

```
const bundle = sigstore.signAttestation(subject)
const serializedBundle = Bundle.toJSON(bundle)
fs.writeFileSync('bundle.sigstore', serializedBundle)
```

src/sigstore.test.ts
```diff
@@ -24,9 +24,30 @@ describe('sign', () => {
   const payload = Buffer.from('Hello, world!');
 
   // Signer output
-  const signedPayload = {
-    base64Signature: 'signature',
-    cert: 'cert',
+  const signedPayload: Bundle = {
+    mediaType: 'application/vnd.dev.sigstore.bundle+json;version=0.1',
+    content: {
+      $case: 'messageSignature',
```
feelepxyz, line 30: TIL: This format output format is set by configuring `ts-proto` with `oneof=unions ` https://github.com/stephenh/ts-proto#oneof-handling

So this is an implementation detail of `ts-proto`, e.g. using buf's `protobuf-es` https://github.com/bufbuild/protobuf-es/blob/main/docs/generated_code.md#oneof-groups yields a slightly different output.

bdehamer:
Closing this in favor of https://github.com/sigstore/sigstore-js/pull/121

PR #113, opened 2022-10-14
Add basic help and version command line option

danbev:
#### Summary
This commit adds very basic command line `help` and `version` command line options.

The motivation for this that I tried running sigstore --help out of habit but that returned `Unknown command`. Perhaps others will do this same and this commit could be a start of adding more options/details.
This can be tested using the following commands:
```console
$ npm run build
$ ./bin/sigstore.js --help
sigstore <command> <artifact>

  Usage:

  sigstore sign         sign an artifact
  sigstore sign-dsse    sign an artifact using dsse (Dead Simple Signing Envelope)
  sigstore verify       verify an artifact
  
```
Perhaps this can be expanded upon and cleaned up by using a module like yargs, but I was not sure if adding a dependency would be considered ok, and if there might be preferences of which module to use in that case.

#### Release Note

None

#### Documentation

None

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -30,9 +30,35 @@ async function cli(args: string[]) {
     case 'verify':
       await verify(args[1], args[2]);
       break;
+    case 'version':
+    case '-version':
+    case '--version':
+    case '-v':
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      console.log(require('../../package.json').version);
+      break;
+    case 'help':
+    case '--help':
+    case '-h':
+    case '-?':
+      printUsage();
+      break;
     default:
       throw 'Unknown command';
   }
+}
+
+function printUsage() {
+  console.log(`sigstore <command> <artifact>
+
+  Usage:
+
+  sigstore sign         sign an artifact
+  sigstore sign-dsse    sign an artifact using dsse (Dead Simple Signing Envelope)
+  sigstore verify       verify an artifact
+  sigstore version      print version information
+  sigstore help         print help information
+  `);
 }
 
 const signOptions = {


src/cli/index.ts
```diff
@@ -30,11 +31,35 @@ async function cli(args: string[]) {
     case 'verify':
       await verify(args[1], args[2]);
       break;
+    case 'version':
+    case '-version':
+    case '--version':
+    case '-v':
+      console.log(packageJson.version);
+    console.log("bajja");
```
bdehamer, line 39: I'm guessing this is something leftover from debugging?

tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "declaration": true,
     "declarationMap": true,
     "noImplicitAny": true,
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "resolveJsonModule": true
```
bdehamer, line 8: We played this option when we were trying to read the package version as part of the [getUserAgent](https://github.com/sigstore/sigstore-js/blob/main/src/util/ua.ts) function. What we observed is that it kinda creates a mess when the package is tar'ed (`npm pack`). Ultimately, we just resorted to [requiring](https://github.com/sigstore/sigstore-js/blob/main/src/util/ua.ts#L22) the package JSON so that we didn't have to use this setting at all.

bdehamer (commented):
Thanks for the contribution! I have some concerns w/ the way the package version is being retrieved, but looks good otherwhise.

src/cli/index.ts
danbev: Sorry about that, I'll remove it shortly. Thanks

tsconfig.json
danbev: Ah I see, I'll change this to user `require`. Thanks

bdehamer (approved):
👍

PR #121, opened 2022-10-20
Serialized bundle types

bdehamer:
#### Summary

Adds TypeScript types for the serialized form of the protobuf-generated Sigstore Bundle format. Updates the public package interface (`sign`, `signAttestation`, & `verify`) to use the new serialized Bundle types in place of the internal Bundle representation.

#### Release Note
* Public sign/verify methods now return/accept serialized form of Sigstore bundle.


src/cli/index.ts
```diff
@@ -49,7 +48,7 @@ async function sign(artifactPath: string) {
   console.error(`Created entry at index ${logIndex}, available at`);
   console.error(`${url}?logIndex=${logIndex}`);
 
-  console.log(JSON.stringify(Bundle.toJSON(bundle)));
+  console.log(JSON.stringify(bundle));
```
bdehamer, line 51: No need for consumers to use `Bundle.toJSON` function as bundle is already in it's serialized form.

src/sigstore.ts
```diff
@@ -44,32 +53,36 @@ type IdentityProviderOptions = Pick<
...
-  return new Signer({
+  const signer = new Signer({
     fulcio,
     rekor,
     identityProviders: idps,
-  }).signBlob(payload);
+  });
+
+  const bundle = await signer.signBlob(payload);
+  return BundleSerializer.toJSON(bundle) as MessageSignatureBundle;
```
bdehamer, line 67: Internal Bundle is translated to it's serialized from immediately before being returned.

src/sigstore.test.ts
```diff
@@ -24,9 +24,8 @@ describe('sign', () => {
   const payload = Buffer.from('Hello, world!');
 
   // Signer output
-  const signedPayload = {
-    base64Signature: 'signature',
-    cert: 'cert',
+  const bundle = {
```
feelepxyz, line 27: Thoughts on using a real example here and on line 77? e.g. https://github.com/sigstore/sigstore-js/pull/107/files#diff-03b4f24fce55af1e22db038dfd3ff8429227f5ca4d6b284b663628b6db4fac2aR27

src/types/bundle/serialized.ts
```diff
@@ -0,0 +1,95 @@
...
+// Serialized form of the Sigstore Bundle union type with all possible options
+// represented
+export type SerializedBundle = {
+  mediaType: string;
+  verificationData: SerializedVerificationData | undefined;
+  verificationMaterial:
+    | Serializedx509CertificateChain
+    | SerializedPublicKey
+    | undefined;
+} & (SerializedMessageSignature | SerializedDSSEEnvelope);
```
feelepxyz, line 79: I had a poke around and found this pattern for [creating a OneOf type](https://www.typescriptlang.org/play?#code/C4TwDgpgBAaghgGwK4QPIDMA8qBGArAPigF4pc8BtAawhAHt0z8BdAbgFgAoUSMgOzR8EIbPgA0UANJQIAD2AQ+AEwDOUGvUbkipAN5RqtKAEs+UAKKyAxsiURMGhkzwTJBZgC4ofJAgRQAXygAMih9QxATM0lPZwoYwI5ucGhUAQwAIRBJWlFCEjCDDSj1WidyWLTBYTyJDSIApLkwOgAnYCgeVPSsABUdWEQUDGwerJyRfoIkri6oAGUIVuNEYwAvCCUAWQgVFTgAcwh54wO+OGAkVug9Lih7qABbXf2jgBFT3eAPO4f7gB8wr8-n9EAc2sZgAALR5eFTAZZ8A5JEEgpSfeFwhGmZHAkEBPEAqBIZQQdCmTYoh4qU7nS7XLGI3GcRpcWYpBZLFYIdabN7zebmcx8ABuEAQdF4t04fzAcBAErgSkZOKp9zlCroSt6KRVSLVUBpZwuV12Xl0hMNpz1zNRGmMysN2P1eICFDYXFZnDZyV4i2Wqw2SgyJKUCBuQJlT02Kx1kBtSSCoSqIwtUeje0Ox1pJoZnIDPKDO0zRxOxvpEDVqhUEGFYol8fz3N5Sn5grr4sllc90x9VjofHhUBwofDXn9zaDIeU4YKab+z3RcDjEC8AHI4GAwDyrBdjAOAPQi5QAOjsIpPRvhbQgJ5HM4gAGo8CoB6wxa0aQPiAAGE8ARjXMQ8WeEts3LU1zUtUDXggD4jkxSNUXuMEIWhWEoDXeYAAkAEEACYAH18IAVgANiAy17nRBDvkw-86BI-8ABkD0kH83nMHAAHctgAMQAFgEn8AEkAGYqHwsiwAOAA5Xotlk7itxEvgwAAJUfOAAFUtJ-bDuOISj0weAJgJM+4jTpSCqMwrZzG0gBhETzAyPgAAUAEUQHcg8+DWKgkAyAAtKwlAADTI9A4G4gAOXowDeABxKhZC2AApR9zGMRyDhAASRNaYKfyoPDjHMXDYrwR8+D4vjZOMESSPS2SxLIkBZJAc46DIpBmO4n9gpIziAE4RqS4xH0eDyAHZPJgR9tKoIzzL+AkWR9YxGAACnvMNbxgrMy2s64AEokPuftBw6KzcwjPbwxPQ7SxzCtL1e00ZnTA8DygWS6A6OArBeYwcHDYEHtvata1FTtIBPDVFSUFdPS4IA). Spotted in [this blog post](https://dev.to/maxime1992/implement-a-generic-oneof-type-with-typescript-22em).

Not sure if this is the best way of representing this but seems like this could mean we can get rid of `SerializedMessageSignature | SerializedDSSEEnvelope` and just expose a `SerializedBundle` interface for all inputs and outputs.

src/sigstore.test.ts
bdehamer: Yeah, good call. This will give me an opportunity to check that the return value is truly the `Bundle.toJSON` form of the bundle.

src/sigstore.test.ts
```diff
@@ -117,13 +195,14 @@ describe('signAttestation', () => {
...
   });
 });
 
 describe('#verify', () => {
   const bundle: SerializedBundle = {
     mediaType: 'application/vnd.dev.sigstore.bundle+json;version=0.1',
+    dsseEnvelope: undefined,
     messageSignature: {
       messageDigest: {
         algorithm: 'SHA2_256',
```
bdehamer, line 208: This is the one bit of the `OneOf` type that I don't like . . . the unused alternatives still need to be present, but set to `undefined`. However, this should really only come up in situations like this where we're manually constructing one of these objects to use in a test.

src/sigstore.ts
```diff
@@ -53,7 +46,7 @@ type IdentityProviderOptions = Pick<
 export async function sign(
   payload: Buffer,
   options: SignOptions = {}
-): Promise<MessageSignatureBundle> {
+): Promise<Bundle> {
```
bdehamer, line 49: Note that `Bundle` here is actually `SerializedBundle`, but I've aliased the type so that we don't expose this internal detail to package consumers.

bdehamer:
@feelepxyz Thanks for the pointer to the `OneOf` utility! I think this looks pretty good now.

I did refactor `OneOf` a bit to use `undefined` instead of `null` for the value of the excluded options. This felt a little safer since it will stringify into something that matches the expected serialization format:

```
> JSON.stringify({a: 'foo', b: null})
'{"a":"foo","b":null}'
> JSON.stringify({a: 'foo', b: undefined})
'{"a":"foo"}'
```

src/types/utility.ts
```diff
@@ -0,0 +1,32 @@
...
+*/
+// https://dev.to/maxime1992/implement-a-generic-oneof-type-with-typescript-22em
+
+// Returns a type that is the union of all value types in the given object.
+type ValueOf<Obj> = Obj[keyof Obj];
+
+// Returns a type narrowing the given object to only the given key -- all
+// other keys must be undefined.
+type OneOnly<Obj, K extends keyof Obj> = {
+  [key in Exclude<keyof Obj, K>]: undefined;
```
feelepxyz, line 24: Nice, undefined makes more sense 👍

src/types/bundle/serialized.test.ts
```diff
@@ -0,0 +1,317 @@
...
+import { Envelope } from './__generated__/envelope';
+import { Bundle, VerificationData } from './__generated__/sigstore_bundle';
+import {
+  HashAlgorithm,
+  hashAlgorithmToJSON,
+  MessageSignature,
+  X509CertificateChain,
+} from './__generated__/sigstore_common';
+
+describe('Serialized Types', () => {
```
feelepxyz, line 26: Appreciate you adding these extensive tests 💯

feelepxyz (approved):
This is much appreciated! This looks great to me.

Maybe one day we can patch real "one of" types into typescript and patch one of the proto to TS generators to also generate these serialized types 😛

PR #128, opened 2022-10-24
Support custom signer callback

bdehamer:

#### Summary

Updates the `Signer` class to support a user-provided `SignerFunc` which is responsible for generating and returning an artifact signature. If no `SignerFunc` is provided, it defaults to using the standard ephemeral keypair w/ Fulcio certificate signing scheme.

Allowing for a a custom signing function allows us to support scenarios where a user may want to use something like a KMS key to generate a signature which still utilizing the rest of the Sigstore workflow (upload to Rekor, bundle creation, etc).

#### Release Note

Lays the foundation for user-supplied signing functions. Note that this isn't yet exposed as part of the package's public interface. Will add that in a subsequent commit.

### DIFF
--- a/src/sign.ts
+++ b/src/sign.ts
@@ -18,22 +18,20 @@ import { Provider } from './identity';
 import { Bundle, bundle, Envelope } from './types/bundle';
 import { fulcio } from './types/fulcio';
 import { rekor } from './types/rekor';
+import { SignatureMaterial, SignerFunc } from './types/signature';
 import { crypto, dsse, oidc, pem } from './util';
 
 export interface SignOptions {
   fulcio: Fulcio;
   rekor: Rekor;
   identityProviders: Provider[];
-}
-
-interface SigCert {
-  signature: Buffer;
-  certificates: string[];
+  signer?: SignerFunc;
 }
 
 export class Signer {
   private fulcio: Fulcio;
   private rekor: Rekor;
+  private signer: SignerFunc;
 
   private identityProviders: Provider[] = [];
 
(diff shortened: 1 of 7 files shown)


src/types/signature.ts
```diff
@@ -0,0 +1,17 @@
...
+    id?: string;
+    value: string;
+  };
+}
+
+// Result of signing some artifact. Containing the signature and either the
+// signing certificate chain or the public key.
+export type SignatureMaterial = {
+  signature: Buffer;
+} & OneOf<VerificationMaterial>;
```
bdehamer, line 15: Introduces a new, internal type to carry around the generated signature and information about the certificate/key used to generate that signature.

src/types/rekor/index.ts
```diff
@@ -96,12 +94,11 @@ export const rekor = {
...
     digest: Buffer,
-    signature: Buffer,
-    certificate: string
+    signature: SignatureMaterial
   ): HashedRekordKind => {
     const b64Digest = digest.toString('hex');
-    const b64Signature = signature.toString('base64');
-    const b64Certificate = enc.base64Encode(certificate);
+    const b64Signature = signature.signature.toString('base64');
+    const b64Key = enc.base64Encode(toPublicKey(signature));
```
feelepxyz, line 101: What format should the public key be in? Is the `publicKey.content` sent to Rekor a base64 encode DER key? In which case, should `signature.key.value` be a `Buffer` as DER should be binary?

feelepxyz: Just had a quick poke in node and looks like you get a Buffer if you export a `PublicKeyObject` with `publicKey.export({format: 'der', type: 'spki'})`

feelepxyz: Does this mess with `certificates` being strings?

feelepxyz (commented):
Thanks for jumping on this, changes look good, just had one question around the public key format.

src/sign.ts
```diff
@@ -75,32 +66,29 @@ export class Signer {
...
     };
 
-    const signingCertificate = certificates[0];
-
-    const proposedEntry = rekor.toProposedIntotoEntry(
-      envelope,
-      signingCertificate
-    );
+    const proposedEntry = rekor.toProposedIntotoEntry(envelope, sigMaterial);
     const entry = await this.rekor.createEntry(proposedEntry);
```
feelepxyz, line 84: Any thoughts on how we could get this proposed entry and not create the Rekor entry at the same time? The interface doesn't feel ideal, as it looks like we'd store the `proposedEntry` and `sigMaterial` to later create the entry and the dsse bundle 🤔

src/types/rekor/index.ts
bdehamer: The `toPublicKey` function here simply extracts either the signing certificate or the key from the `SignatureMaterial` object. In both cases the values should be PEM strings which Rekor wants base64-encoded.

src/types/signature.ts
```diff
@@ -0,0 +1,17 @@
+import { OneOf } from './utility';
+
+interface VerificationMaterial {
+  certificates: string[];
+  key: {
+    id?: string;
+    value: string;
```
feelepxyz, line 7: Should we add a comment here to say it's pem encoded?

feelepxyz (approved):

PR #133, opened 2022-10-25
Expose new addEnvelopToTransparencyLog function

bdehamer:

#### Summary

Exposes a new public `addEnvelopeToTransparencyLog` function which allows a user with an existing, signed DSSE envelope to post that envelope to Rekor and receive a Sigstore bundle in return.

```typescript
function addEnvelopeToTransparencyLog(envelope: Envelope, key: string): Promise<Bundle> { }
```

As part of exposing this function, I introduced a new `TLog` class which sits atop the `Rekor` class. Whereas `Rekor` provided direct access to the Rekor REST API, the new `TLog` class encapsulates the logic necessary to marshal data to/from the Rekor request/response formats. Ultimately, I'd like to do something similar on top of the `Fulcio` client, but I'll save that for another refactor.
#### Release Note

* Adds new `addEnvelopeToTransparencyLog` to public interface.

### DIFF
--- a/src/client/rekor.ts
+++ b/src/client/rekor.ts
@@ -18,11 +18,9 @@ import { Entry, EntryKind } from '../types/rekor';
 import { ua } from '../util';
 import { checkStatus } from './error';
 
-const DEFAULT_BASE_URL = 'https://rekor.sigstore.dev';
-
 // Client options
 export interface RekorOptions {
-  baseURL?: string;
+  baseURL: string;
 }
 
 export interface SearchIndex {
@@ -52,11 +50,8 @@ export class Rekor {
         'User-Agent': ua.getUserAgent(),
       },
     });
-    this.baseUrl = Rekor.getBaseUrl(options.baseURL);
-  }
 
-  public static getBaseUrl(baseURL?: string) {
-    return baseURL ?? DEFAULT_BASE_URL;
+    this.baseUrl = options.baseURL;
   }
 
   /**
--- a/src/sign.ts
+++ b/src/sign.ts
@@ -13,31 +13,31 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import { Fulcio, Rekor } from './client';
+import { Fulcio } from './client';
 import { Provider } from './identity';
-import { Bundle, bundle, Envelope } from './types/bundle';
+import { TLog } from './tlog';
+import { Bundle, Envelope } from './types/bundle';
 import { fulcio } from './types/fulcio';
-import { rekor } from './types/rekor';
 import { SignatureMaterial, SignerFunc } from './types/signature';
 import { crypto, dsse, oidc, pem } from './util';
 
 export interface SignOptions {
   fulcio: Fulcio;
-  rekor: Rekor;
+  tlog: TLog;
   identityProviders: Provider[];
   signer?: SignerFunc;
 }
 
 export class Signer {
   private fulcio: Fulcio;
-  private rekor: Rekor;
+  private tlog: TLog;
   private signer: SignerFunc;
 
   private identityProviders: Provider[] = [];
 
   constructor(options: SignOptions) {
     this.fulcio = options.fulcio;
-    this.rekor = options.rekor;
+    this.tlog = options.tlog;
     this.identityProviders = options.identityProviders;
     this.signer = options.signer || this.signWithEphemeralKey.bind(this);
   }
@@ -50,13 +50,7 @@ export class Signer {
     const digest = crypto.hash(payload);
 
     // Create Rekor entry
-    const proposedEntry = rekor.toProposedHashedRekordEntry(
-      digest,
-      sigMaterial
-    );
-    const entry = await this.rekor.createEntry(proposedEntry);
-
-    return bundle.toMessageSignatureBundle(digest, sigMaterial, entry);
+    return this.tlog.createMessageSignatureEntry(digest, sigMaterial);
   }
 
   public async signAttestation(
@@ -80,10 +74,7 @@ export class Signer {
       ],
     };
 
-    const proposedEntry = rekor.toProposedIntotoEntry(envelope, sigMaterial);
-    const entry = await this.rekor.createEntry(proposedEntry);
-
-    return bundle.toDSSEBundle(envelope, sigMaterial, entry);
+    return this.tlog.createDSSEEntry(envelope, sigMaterial);
   }
 
   private async signWithEphemeralKey(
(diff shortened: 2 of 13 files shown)


src/client/rekor.ts
```diff
@@ -52,11 +50,8 @@ export class Rekor {
         'User-Agent': ua.getUserAgent(),
       },
     });
-    this.baseUrl = Rekor.getBaseUrl(options.baseURL);
-  }
 
-  public static getBaseUrl(baseURL?: string) {
-    return baseURL ?? DEFAULT_BASE_URL;
+    this.baseUrl = options.baseURL;
```
bdehamer, line 54: Moved the default URL higher in the stack so that we have fewer optional parameters to juggle.

src/sign.ts
```diff
@@ -50,13 +50,7 @@ export class Signer {
...
 
     // Create Rekor entry
-    const proposedEntry = rekor.toProposedHashedRekordEntry(
-      digest,
-      sigMaterial
-    );
-    const entry = await this.rekor.createEntry(proposedEntry);
-
-    return bundle.toMessageSignatureBundle(digest, sigMaterial, entry);
+    return this.tlog.createMessageSignatureEntry(digest, sigMaterial);
```
bdehamer, line 53: All of the data transformation necessary to post to Rekor and format the response as a Bundle is now encapsulated behind the `TLog` interface.

src/sigstore.ts
```diff
@@ -66,30 +78,42 @@ export async function signAttestation(
...
+export async function addEnvelopeToTransparencyLog(
+  envelope: Envelope,
+  key: string,
+  options: SignOptions = {}
+): Promise<Bundle> {
+  const tlog = newTLogClient(options);
+  const bundle = await tlog.createDSSEEntry(envelope, key);
+  return bundleToJSON(bundle) as Bundle;
+}
+
```
bdehamer, line 116: Encapsulating all the Rekor logic within the `TLog` class means we can easily reuse the implementation both here and in the Signer.

bdehamer: Open to suggestions for naming this function. Don't love what I've come up with here but struggled to find something better.

src/sigstore.ts
```diff
@@ -43,21 +49,27 @@ type IdentityProviderOptions = Pick<
   'identityToken' | 'oidcIssuer' | 'oidcClientID' | 'oidcClientSecret'
 >;
 
+function newTLogClient(options: { rekorBaseURL?: string }): TLog {
```
feelepxyz, line 52: prefixing with `new` feels weird in js, I initially thought it was a typo and should be `new TLogClient` on line 63. 

Thoughts on just down casing the instance?
```suggestion
function tLogClient(options: { rekorBaseURL?: string }): TLog {
```

src/sigstore.ts
```diff
@@ -43,21 +49,27 @@ type IdentityProviderOptions = Pick<
...
+  });
+}
+
 export async function sign(
   payload: Buffer,
   options: SignOptions = {}
 ): Promise<Bundle> {
   const fulcio = new Fulcio({ baseURL: options.fulcioBaseURL });
-  const rekor = new Rekor({ baseURL: options.rekorBaseURL });
+  const tlog = newTLogClient(options);
```
feelepxyz, line 63: this diff broke my brain :)

src/sigstore.ts
feelepxyz: Yeah not sure I have many better ideas but any reason for not calling out rekor here, e.g:
- `createRekorEntryFromDsseEnvelope`
- addDsseEnvelopeToRekor`

src/sigstore.ts
feelepxyz: oh wait it's returning a "new tlog client".. maybe
```suggestion
function createTLogClient(options: { rekorBaseURL?: string }): TLog {
```

src/tlog/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    return bundle.toMessageSignatureBundle(digest, sigMaterial, entry);
+  }
+
+  async createDSSEEntry(
+    envelope: Envelope,
+    sigMaterial: SignatureMaterial | string
+  ): Promise<Bundle> {
+    // If sigMaterial is a string, assume it's a PEM-encoded public key
+    // and create a new SignatureMaterial object from data in envelope
+    if (typeof sigMaterial === 'string') {
```
feelepxyz, line 63: This string check definitely feels a bit surprising here, thoughts on pushing this logic up a level to `addEnvelopeToTransparencyLog` where you know it's a public key?

feelepxyz (commented):
Appreciate you jumping on this so quick 🙇  Added some minor comments otherwise LGTM!

src/sigstore.ts
```diff
@@ -66,30 +78,42 @@ export async function signAttestation(
...
-  const b = BundleSerializer.fromJSON(bundle);
+  const b = bundleFromJSON(bundle);
   return verifier.verify(b, data);
 }
 
+// Accepts a signed DSSE envelope and a PEM-encoded public key to be added to the
+// transparency log. Returns a Sigstore bundle suitable for offline verification.
+export async function addEnvelopeToTransparencyLog(
+  envelope: Envelope,
+  key: string,
```
feelepxyz, line 109: Any reason we wouldn't name these `dsseEnvelope` and `publicKey` to match the stuff in bunde/rekor?
```suggestion
  dsseEnvelope: Envelope,
  publicKey: string,
```

src/tlog/index.test.ts
```diff
@@ -0,0 +1,325 @@
...
+        // Mock Rekor request
+        nock(baseURL)
+          .matchHeader('Accept', 'application/json')
+          .matchHeader('Content-Type', 'application/json')
+          .post('/api/v1/log/entries', proposedEntry)
+          .reply(201, rekorEntry);
+      });
+
+      it('returns a signature bundle', async () => {
+        const bundle = await subject.createDSSEEntry(dsse, sigMaterial);
```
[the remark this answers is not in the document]
bdehamer: Yeah, I definitely find myself writing a lot of the same test code over-and-over (luckily, copilot has been doing much of the heavy lifting). I have some ideas for a big refactor of the unit tests which will set-up some reusable fixtures and DRY up some of the expectations. Maybe once we get past this next milestone I'll do a pass on this.

src/sigstore.ts
bdehamer: I sorta had it in my head that "Rekor" was an implementation detail that we didn't want to expose . . . but I'm probably overthinking it. This is the SIGSTORE library after all.

I think I'll go with `createRekorEntry`

src/sigstore.ts
bdehamer: 👍

src/sigstore.ts
bdehamer: 👍

src/tlog/index.ts
bdehamer: Yeah, good call. I extracted this bit of logic into a reusable function and pushed it up a level.

feelepxyz (approved):
LGTM!

PR #136, opened 2022-10-28
Accept serialized envelope in createRekorEntry

feelepxyz:
Modifies the interface of `createRekorEntry` to take a "serialized dsse envelope" which is a JSON representation and de-serialize it internally.

### DIFF
--- a/src/types/bundle/serialized.ts
+++ b/src/types/bundle/serialized.ts
@@ -83,3 +83,15 @@ export type SerializedBundle = {
   dsseEnvelope: SerializedDSSEEnvelope;
   messageSignature: SerializedMessageSignature;
 }>;
+
+interface SerializedSignature {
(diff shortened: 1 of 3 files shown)


src/types/bundle/serialized.ts
```diff
@@ -23,24 +23,24 @@ type SerializedVerificationData = {
       keyId: string;
     };
     kindVersion:
-      | {
-          kind: string;
-          version: string;
-        }
-      | undefined;
+    | {
```
feelepxyz, line 26: hm not sure what happened here, I think `prettier` just ran. Should this be set in config somewhere?

bdehamer (approved):
Nice! :shipit:

Shawnblake (commented):
Nice

PR #137, opened 2022-10-29
rekor set verification

bdehamer:

#### Summary
Updates the `verify` function to perform offline verification of the Rekor signed entry timestamp (SET) for each tlog entry present in the Sigstore bundle.

To test:

```bash
# Create some artifact to sign
cat -n "hello world" > artifact

# Sign artifact
./bin/sigstore.js sign artifact > bundle.sigstore

# Verify both the artifact signature and the Rekor SET in the bundle
./bin/sigstore.js verify bundle.sigstore artifact
```
#### Release Note
* `verify` performed offline Rekor SET verification in addition to artifact signature verification.


src/cli/index.ts
```diff
@@ -101,12 +101,15 @@ async function verify(bundlePath: string, artifactPath: string) {
...
-
-  if (result) {
+  try {
+    await sigstore.verify(bundle, payload);
     console.error('Verified OK');
-  } else {
-    throw 'Signature verification failed';
+  } catch (e) {
+    console.error('Verification failed');
+    if (e instanceof Error) {
```
feelepxyz, line 109: Aren't all error's instances of the `Error` class? Or what is this guard protecting against?

src/tlog/verify.ts
```diff
@@ -0,0 +1,169 @@
...
+
+function toPublicKey(verificationMaterial: VerificationMaterial): string {
+  switch (verificationMaterial?.content?.$case) {
+    case 'x509CertificateChain': {
+      const der =
+        verificationMaterial.content.x509CertificateChain.certificates[0];
+      return pem.fromDER(der.derBytes);
+    }
+    case 'publicKey':
+    // TODO: How to handle this?
```
feelepxyz, line 126: Do we need to figure this out for verifying the SET of publish attestation?

src/tlog/verify.ts
```diff
@@ -0,0 +1,169 @@
...
+  const sigMaterial: SignatureMaterial = {
+    signature: sig,
+    certificates: [certificate],
+    key: undefined,
+  };
+
+  const body = toProposedIntotoEntry(dsseEnvelope, sigMaterial);
+
+  // When Rekor saves the entry it removes the payload from the envelope
+  if (body.apiVersion === '0.0.2') {
```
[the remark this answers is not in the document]
bdehamer: Yep! This but in particular (SET verification) seems especially brittle.

src/cli/index.ts
bdehamer: In JS you can technically `throw` anything, so if I want to access `message` here TypeScript forces me to first prove it's an `Error`.

feelepxyz (approved):
Awesome work on this! I bet there was a lot of head bashing getting the payload in the correct shape for verification to pass 🎉 

I see the verify command now explodes if it doesn't verify, this seems good to me as it doesn't differentiate between errors caused by input or implementation, both of which should always hard fail verification.

src/tlog/verify.ts
bdehamer: Yeah. I'll want to see an example of one of those entries in Rekor in order to figure out how this should work.

feelepxyz: Here's an example: https://rekor.sigstore.dev/api/v1/log/entries?logIndex=5983279
and here's the bundle: https://github.com/github/package-security/issues/439

PR #138, opened 2022-10-31
Add createDSSEEnvelope method

feelepxyz:
Creates a dsse envelope using a custom signer. The output of this function can be passed to `createRekorEntry` along with a public key in order to separate signing and the Rekor upload.


bdehamer (approved):
Nice! I think this is a good approach, and one we can continue to use if we find the need to export other helper functions.

PR #139, opened 2022-10-31
Reduce Dependabot PRs

feelepxyz:
This should reduce the number of Dependabot PRs but no longer submitting PRs that only change the lockfile.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ updates:
   directory: "/"
   schedule:
     interval: daily
-  versioning-strategy: auto
+  versioning-strategy: increase-if-necessary
   open-pull-requests-limit: 10
 - package-ecosystem: "github-actions"
   directory: "/"


.github/dependabot.yml
```diff
@@ -4,7 +4,7 @@ updates:
   directory: "/"
   schedule:
     interval: daily
-  versioning-strategy: auto
+  versioning-strategy: increase-if-necessary
```
feelepxyz, line 7: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

feelepxyz: hm not entirely sure about this, but dunno why we are seeing all these lockfile only PRs https://github.com/sigstore/sigstore-js/pull/135/files

feelepxyz: @bdehamer oh actually, we should just delete/gitignore `package-lock.json` as it doesn't apply when you `npm install` it.

feelepxyz: it's all coming back to me.. package-lock is useful for keeping dev-dependencies locked but not for library deps as these just get pulled from `package.json`. I feel like dependabot didn't used create lockfile only updates unless it was a security update 🤔

feelepxyz:
Don't think this will do anything :(

PR #142, opened 2022-10-31
verify tlog integrated time against signing cert

bdehamer:

#### Summary
Updates the verification logic with a check which will compare the `integratedTime` of all tlog entries against the validity period of the signing certificate. 

If the `integratedTime` occurs before the certificate's "Not Before" time or after the certificate's "Not After" time the bundle will be considered invalid.


feelepxyz (approved):
LGTM 👍

PR #147, opened 2022-11-01
Fix absolute path to rekor public key

feelepxyz:
When install `sigstore` as a library the library is executed from the projects root, so reading `'store/rekor.pub'` is not found and errors: `ENOENT: no such file or directory, open 'store/rekor.pub'`


bdehamer (approved):
Thanks for fixing this!

PR #151, opened 2022-11-03
update rekor intoto/0.0.2 entry hash calculation

bdehamer:

#### Summary

Updates the logic to properly calculate the digest of a DSSE envelope when creating an intoto/0.0.2 style entry in Rekor.

### DIFF
--- a/src/tlog/format.ts
+++ b/src/tlog/format.ts
@@ -75,14 +75,22 @@ function toProposedIntotoV002Entry(
   envelope: Envelope,
   signature: SignatureMaterial
 ): IntotoKind {
+  // Calculate the value for the payloadHash field in the Rekor entry
+  const payloadHash = crypto.hash(envelope.payload).toString('hex');
+
+  // Calculate the value for the hash field in the Rekor entry
+  const envelopeHash = calculateDSSEHash(envelope);
+
+  // Collect values for re-creating the DSSE envelope.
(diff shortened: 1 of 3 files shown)


src/tlog/format.ts
```diff
@@ -111,6 +117,28 @@ function toProposedIntotoV002Entry(
   };
 }
 
+// Calculates the hash of a DSSE envelope for inclusion in a Rekor entry.
+// There is no standard way to do this, so the scheme we're using as as
+// follows:
+//  * payload is base64 encoded
+//  * signature is base64 encoded (only the first signature is used)
+//  * keyid is included ONLY if it is NOT an empty string
+//  * The resulting JSON is canonicalized and hashed to a hex string
```
feelepxyz, line 126: Yikes!

feelepxyz: Is it worth writing some kind of tests around this behaviour to make sure we don't deviate from it in future?

feelepxyz: Do we have some tests in src/tlog/verify.test.ts that create different kinds of envelopes with and without a keyid? Maybe with multiple sigs?

feelepxyz (commented):
Gnarly.. nice work figuring this out! Just had one comment around tests, otherwise LGTM 👍

codysoyland (dismissed):
🎉

feelepxyz (approved):
Thanks for adding the tests!

PR #152, opened 2022-11-04
support for public key retrieval callback

bdehamer:
#### Summary
Updates the `verify` function to accept an optional `findKey` callback which can be used to retrieve the public key for verifying signatures which are not using a signing certificate (cases where verification material in the bundle specifies a public key hint instead of a certificate chain).

In a subsequent PR, the same key returned by `findKey` will also be passed-down to the tlog verification logic and used to re-create the Rekor entry for SET verification.

#### Release Note
* Updates the signature for the `verify` function to accept an optional `findKey` callback which will be invoked to retrieve the public key in scenarios where the signature was generated w/ a static keypair.

### DIFF
--- a/src/sigstore.ts
+++ b/src/sigstore.ts
@@ -44,7 +44,7 @@ export type SignOptions = {
 } & TLogOptions;
 
 export type VerifierOptions = {
-  findKey?: FindKeyFunc;
+  getPublicKey?: FindKeyFunc;
 } & TLogOptions;
 
 export type Bundle = SerializedBundle;
@@ -104,7 +104,11 @@ export async function verify(
 ): Promise<void> {
   const tlog = createTLogClient(options);
   const tlogKeys = getKeys();
-  const verifier = new Verifier({ tlog, tlogKeys, findKey: options.findKey });
+  const verifier = new Verifier({
+    tlog,
+    tlogKeys,
+    getPublicKey: options.getPublicKey,
+  });
 
   const b = bundleFromJSON(bundle);
   return verifier.verifyOffline(b, data);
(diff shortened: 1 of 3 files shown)


src/sigstore.ts
```diff
@@ -43,7 +43,9 @@ export type SignOptions = {
   oidcClientSecret?: string;
 } & TLogOptions;
 
-export type VerifierOptions = TLogOptions;
+export type VerifierOptions = {
+  findKey?: FindKeyFunc;
```
feelepxyz, line 47: this is a very small nitpick, `find` sounds somehow out of place.. not entirely sure why, but wondering if `getKey` sounds better, or `getPublicKey` even? Like from the libs perspective, it's getting the key not finding it?

feelepxyz (dismissed):
Thanks for adding this! 👍  had a very small nitpick on naming

src/sigstore.ts
bdehamer: Good call. Updated!

src/sigstore.ts
```diff
@@ -43,7 +43,9 @@ export type SignOptions = {
   oidcClientSecret?: string;
 } & TLogOptions;
 
-export type VerifierOptions = TLogOptions;
+export type VerifierOptions = {
+  getPublicKey?: FindKeyFunc;
```
feelepxyz, line 47: 😬 

```suggestion
  getPublicKey?: GetPublicKeyFunc;
```

bdehamer: 🤦🏻‍♂️

feelepxyz (approved):

PR #153, opened 2022-11-04
refactor key handling for rekor SET verification

bdehamer:
#### Summary
Continues the work in https://github.com/sigstore/sigstore-js/pull/152 to integrate a `getPublicKey` callback option which can be used for verification when artifacts are signed w/ a static keypair. This change passes the retrieved key down into the SET verification function to ensure that the correct key value is used when re-creating the Rekor entry.


feelepxyz (approved):
A great plan comes together 😄

PR #158, opened 2022-11-09
handle new canonicalizedBody in bundle

bdehamer:
#### Summary
Updates to the latest bundle format w/ the embedded, canonical Rekor entry.

This change to the bundle format has a few implications for bundle verification which are addressed in this PR:
* With the canonical Rekor entry, SET verification no longer depends on re-creaing the original Rekor entry. The logic to verify the Rekor SET has been significantly simplified.
* In order to trust the canonical Rekor entry for SET verification we must now verify that the data contained in the canonical entry actually matches the rest of the data in the bundle. New logic has been introduced to parse the canonical entry and match it against the artifact/DSSE signatures embedded in the bundle.

This PR also introduces a set of test bundle fixtures that we can use to exercise different verification scenarios. Ultimately, we'll use these bundles across all of the different test cases, but for now they're only being used for these Rekor verification tests.


src/tlog/verify.ts
```diff
@@ -31,12 +34,11 @@ import {
 // key.
 export function verifyTLogSET(
   bundle: Bundle,
-  publicKey: string,
```
bdehamer, line 34: Note that we no longer need to pass-in the signing key since we don't have to re-create the original Rekor entry.

src/tlog/verify.ts
```diff
@@ -85,89 +87,152 @@ export function verifyTLogIntegratedTime(bundle: Bundle): void {
...
-  const body = toVerificationBody(bundle, publicKey);
+  switch (body.kind) {
+    case 'intoto':
+      verifyIntotoTLogBody(body, bundle);
+      break;
+    case 'hashedrekord':
+      verifyHashedRekordTLogBody(body, bundle);
+      break;
+    default:
+      throw new Error('Unknown kind found in tlog entry');
```
feelepxyz, line 117: Maybe error the kind here?
```suggestion
      throw new Error(`Unsupported kind found in tlog entry: ${body.kind}`);
```

src/tlog/verify.ts
```diff
@@ -85,89 +87,152 @@ export function verifyTLogIntegratedTime(bundle: Bundle): void {
...
-  return {
-    body: enc.base64Encode(json.canonicalize(body)),
-    integratedTime: Number(integratedTime),
-    logIndex: Number(logIndex),
-    logID: logId.keyId.toString('hex'),
-  };
+// Compare the given intoto tlog entry to the given bundle
+function verifyIntotoTLogBody(tlogEntry: IntotoKind, bundle: Bundle): void {
+  if (bundle.content?.$case !== 'dsseEnvelope') {
+    throw new Error('bundle content and tlog entry do not match');
```
feelepxyz, line 124: Should this error message say something else?
```suggestion
    throw new Error(`Unsupported bundle content: ${bundle.content?.$case || 'unknown'}`);
```

src/tlog/verify.ts
```diff
@@ -85,89 +87,152 @@ export function verifyTLogIntegratedTime(bundle: Bundle): void {
...
-    case 'dsseEnvelope': {
-      return toDSSEVerificationBody(bundle.content.dsseEnvelope, publicKey);
-    }
+// Compare the given hashedrekord tlog entry to the given bundle
+function verifyHashedRekordTLogBody(
+  tlogEntry: HashedRekordKind,
+  bundle: Bundle
+): void {
+  if (bundle.content?.$case !== 'messageSignature') {
+    throw new Error('bundle content and tlog entry do not match');
```
feelepxyz, line 144:
```suggestion
    throw new Error(`Unsupported bundle content: ${bundle.content?.$case || 'unknown'}`);
```

src/tlog/verify.ts
```diff
@@ -85,89 +87,152 @@ export function verifyTLogIntegratedTime(bundle: Bundle): void {
...
   }
 
-  // Rekor metadata
-  const { integratedTime, logIndex, logId } = entries[index];
+  const { kind, version } = entry.kindVersion;
+  const body: EntryKind = JSON.parse(entry.canonicalizedBody.toString('utf8'));
 
-  if (!logId) {
-    throw new Error('No logId found in bundle');
+  if (kind !== body.kind || version !== body.apiVersion) {
```
feelepxyz, line 105: All these consistency checks also don't feel great but better than the non-standard SET hash calculation.. trust is hard! 😅

feelepxyz (approved):
LGTM! Minor comments around error messages 💯

PR #161, opened 2022-11-13
refactor tlog verification

bdehamer:

#### Summary
Internal refactoring of the transparency log verification logic to improve efficiency.

Previously, there were three separate transparency log verification steps which were exposed as separate functions:

* Verification of the canonicalized transparency log body
* Verification of the transparency log SET
* Verification of the transparency log integrated time.

With this refactor, all three of these checks are rolled into a single `verifyTLogEntries` function which performs these operations across all transparency log entries in the bundle.

### DIFF
--- a/src/verify.ts
+++ b/src/verify.ts
@@ -16,11 +16,7 @@ limitations under the License.
 import { KeyObject } from 'crypto';
 import { InvalidBundleError, VerificationError } from './error';
 import { TLog } from './tlog';
-import {
-  verifyTLogBodies,
-  verifyTLogIntegratedTime,
-  verifyTLogSET,
-} from './tlog/verify';
+import { verifyTLogEntries } from './tlog/verify';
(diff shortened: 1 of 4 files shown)


src/verify.ts
```diff
@@ -53,9 +49,7 @@ export class Verifier {
     const publicKey = await this.getPublicKey(bundle);
 
     verifyArtifactSignature(bundle, publicKey, data);
-    verifyTLogSET(bundle, this.tlogKeys);
-    verifyTLogBodies(bundle);
-    verifyTLogIntegratedTime(bundle);
+    verifyTLogEntries(bundle, this.tlogKeys);
```
bdehamer, line 52: This is the primary purpose of this PR -- encapsulate these three verification steps behind a single function.

feelepxyz (approved):
LGTM 💯

PR #181, opened 2022-12-01
fix packaging issue w/ rekor types

bdehamer:
#### Summary
The codegen'ed Rekor types aren't being included in the build output directory and are missing from the published package.

Originally, the types were being generated with a `.d.ts` suffix, but the TypeScript compiler ignores these and doesn't copy them to the output `dist` directory. Changing the suffix to `.ts` ensures that TypeScript will include them as part of the build output (see [this](https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build))


eddiezane (approved):
/lgtm

PR #187, opened 2022-12-07
x509 cert parsing

bdehamer:
#### Summary

Adds support for basic parsing of DER-encoded x509 certificates.


src/__tests__/x509/cert.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        const cert = x509Certificate.fromDER(derCert);
+
+        expect(cert.version).toBe('v3');
+        expect(cert.notBefore).toBeInstanceOf(Date);
+        expect(cert.notBefore.toISOString()).toBe('2021-10-07T13:56:59.000Z');
+        expect(cert.notAfter).toBeInstanceOf(Date);
+        expect(cert.notAfter.toISOString()).toBe('2031-10-05T13:56:58.000Z');
+
+        expect(cert.subject).toHaveLength(42);
+        expect(cert.issuer).toHaveLength(42);
```
feelepxyz, line 21: Any reason to not check the string value or is it not set? 42 feels a bit cryptic  😀

feelepxyz: Or do these lengths mean something?

src/__tests__/x509/cert.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        expect(cert.extKeyUsage?.keyCertSign).toBe(false);
+        expect(cert.extKeyUsage?.crlSign).toBe(false);
+
+        expect(cert.extSCT).toBeDefined();
+        expect(cert.extSCT?.critical).toBe(false);
+      });
+    });
+  });
+
+  describe('verification', () => {
```
feelepxyz, line 98: Very nice seeing this!

feelepxyz (approved):
Awesome! This looks good to me 👍

src/__tests__/x509/cert.test.ts
bdehamer: The `subject` and `issuer` fields are themselves big, nested data structures which we could parse . . . but for purposes of verification we only care that the _hash_ of the parent's `subject` matches the _hash_ of the child's `issuer` so I'm just returning the raw binary for these fields.

For these tests, I just wanted to ensure that these fields contain some data but don't really care what's in it. Perhaps I should do something like "length > 0" so that it doesn't feel quite so magic-numbery.

PR #188, opened 2022-12-07
publish package w/ provenance

bdehamer:
#### Summary

Generate and publish an intoto build provenance statement for the `sigstore` package. At the moment, provenance generation is not publicly available and requires a pre-release version of the npm CLI as well as special configuration on the npm registry.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,10 @@ on:
   release:
     types: [published]
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -17,9 +17,20 @@ jobs:
         node-version: 16
```
feelepxyz, line 17: I think we also need the id token permissions, could add this top-level:

```
permissions:
  contents: read
  id-token: write
```

feelepxyz (approved):

PR #198, opened 2022-12-12
Subject Alternative Name support for x509 Certs

bdehamer:
#### Summary
Updates the `x509Certificate` class with support for the subject-alternate-name extension.


feelepxyz (approved):
LGTM 👍

asraa (commented):
Thanks @bdehamer ! Sorry for the late review but just updating to let you know that I think this is sufficient -- we are currently only extracting the SAN from the cert. this is great!
cc @laurentsimon we can replace https://github.com/slsa-framework/slsa-github-generator/blob/77395259ff7f705e2578f4552b79d90c06f49554/.github/actions/verify-token/src/main.ts#L195 with the x509 util library here

PR #200, opened 2022-12-13
add verify method to x509Certificate class

bdehamer:

#### Summary
Add's a `verify` method to the `x509Certificate` class which can be used to verify the certificate's signature against the public key of its issuer.


feelepxyz (approved):
🎉

Issue #215, opened 2022-12-20
Get TLOG keys from sigstore-js when usedd as a library


asraa:
https://github.com/sigstore/sigstore-js/blob/825f81ca3d4453cf1252f08d63ee6a164eca9b40/src/tlog/keys.ts#L26

I think that due to this line, when invoking verify outside the CLI with sigstore as a library, this attempts to fetch the rekor.pub from the caller's dir:
`
Error: ENOENT: no such file or directory, open '/home/runner/work/_actions/laurentsimon/slsa-delegator/main/actions/store/rekor.pub'
`

Is there some way to reference the module path? I think I can just get around this by supplying my custom path into my node_modules, but just wanted to flag.

laurentsimon:
using `const pem = fs_1.default.readFileSync(path_1.default.join(__dirname, '../../store/rekor.pub'), 'utf-8');` fixes the path problem if you use `node_modules/` for your dependencies, but fails if using [ncc](https://github.com/vercel/ncc) to compile an application - which is encouraged for GH Actions.

Maybe the rekor.pub should be stored in a variable instead?

PR #216, opened 2022-12-20
fix: make rekor.pub a variable

laurentsimon:
Signed-off-by: laurentsimon <laurentsimon@google.com>

closes https://github.com/sigstore/sigstore-js/issues/215

#### Summary
The rekor.pub key is stored as a file. The file path does not resolve properly if the caller calls the API (vs CLI). When using [ncc](https://github.com/vercel/ncc), the file is not packaged with the final .js file. Using a variable solves these problems. Another option seems to use `require.resolve()`; which is the fix we agreed on. 

This is probably a temporary fix until TUF is used.


src/tlog/keys.ts
```diff
@@ -14,18 +14,18 @@ See the License for the specific language governing permissions and
...
   // TODO: This should be be loaded via TUF
-  const pem = fs.readFileSync(
-    path.resolve(__dirname, '../../store/rekor.pub'),
-    'utf-8'
-  );
+  const pem = `
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE2G2Y+2tabdTV5BcGiBIx0a9fAFwr
+kBbmLSGtks4L3qX6yYY0zufBnhC8Ur/iy55GhWP/9A/bY2LhC30M9+RYtw==
+-----END PUBLIC KEY-----`
```
feelepxyz, line 27: This file should in theory be relocated by `ncc` > [webpack-asset-relocator-loader](https://github.com/vercel/webpack-asset-relocator-loader) but I had a go at this and looks like it doesn't like the typescript compiled code here as it's trying to pattern match on `path.*` and TS generates this: `path_1.default.resolve`.

What does seem to work is changing this to use `require.resolve`:

```suggestion
  const pem = fs.readFileSync(
    require.resolve('../../store/rekor.pub'),
    'utf-8'
  );
```

src/tlog/keys.ts
```diff
@@ -14,18 +14,18 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 import { createPublicKey, KeyObject } from 'crypto';
```
feelepxyz, line 16:
```suggestion
import { createPublicKey, KeyObject } from 'crypto';
import fs from 'fs';
```

laurentsimon:
Thanks @feelepxyz! I've accepted your changes.

feelepxyz (approved):
Appreciate the fix!

bdehamer (changes requested):
I think the `rekor.pub` file needs to be added back in now.

laurentsimon:
I have added the rekor.pub file back in. PTAL

feelepxyz (approved):

bdehamer (approved):
👍

PR #219, opened 2022-12-21
Certificate chain verification

bdehamer:

#### Summary
Adds a new function which performs basic x509 certificate chain verification:

* Assembles a chain of certificates from the supplied leaf to the root.
  * The issuer of each cert is found either my matching the `authorityKeyIdentifer` extension to the parent's `subjectKeyIdentifier` function, or by matching the the `issuer` field to the parent's `subject`.
  * The signature of each cert is verified against the public key of the parent.
* Checks that the chain contains at least one certificate identified as a trusted certificate.
* Checks that all of the certificates were valid at the specified date.
* Checks that all non-leaf certificates contain a `basicConstraints` extension marking them as certificate authorities.

This function will be used to verify the chain of certificates present in a Sigstore bundle against of set of trusted CA certificates retrieved from the Sigstore TUF root.


src/__tests__/x509/cert.test.ts
```diff
@@ -32,36 +30,29 @@ describe('x509Certificate', () => {
         expect(cert.extKeyUsage?.keyCertSign).toBe(true);
         expect(cert.extKeyUsage?.crlSign).toBe(true);
 
-        expect(cert.extAuthorityKeyID).toBeDefined();
```
feelepxyz, line 35: How come this is now undefined?

src/x509/verify.ts
```diff
@@ -0,0 +1,202 @@
...
+  private certs: x509Certificate[];
+  private trustedCerts: x509Certificate[];
+  private localCerts: x509Certificate[];
+  private checkDate: Date;
+
+  constructor(opts: VerifyCertificateChainOptions) {
+    this.certs = opts.certs;
+    this.trustedCerts = opts.trustedCerts;
+    this.localCerts = dedupeCertificates([...opts.trustedCerts, ...opts.certs]);
+    this.checkDate = opts.checkDate || new Date();
```
feelepxyz, line 27: It wasn't immediately obvious to me what `checkDate` refers to, wondering if something like `validAt` would be clearer?

src/x509/verify.ts
```diff
@@ -0,0 +1,202 @@
...
+      }
+    });
+
+    return issuers;
+  }
+
+  private checkPath(path: x509Certificate[]): void {
+    if (path.length < 2) {
+      throw new CertificateChainVerificationError(
+        'Certificate chain must contain at least two certificates'
```
feelepxyz, line 157: Worth adding a test case for this?

feelepxyz (commented):
Awesome work on this! I added some minor comments about tests and variable names, otherwise LGTM! 👍

src/__tests__/x509/cert.test.ts
bdehamer: The new cert I generated for testing is intentionally not using an `authorityKeyID` extension so that I cant test different mechanisms for chaining certs together (comparing the issuer and subject of two certs is another way to locate the parent of a given certificate).

src/x509/verify.ts
bdehamer: Given the way the certificate chain is assembled in the code above, it's actually not possible to arrive at this point with fewer than two certificates in the chain (other errors would be thrown before you arrive here). I considered removing this check, but I sorta like that it makes explicit a requirement that is difficult to see otherwise (I was thinking about it from the perspective of someone auditing this code).

Happy to delete if you think we should ditch it.

src/x509/verify.ts
```diff
@@ -0,0 +1,202 @@
...
+        'Intermediate certificate is not a CA'
+      );
+    }
+
+    // Certificate's issuer must match the subject of the next certificate
+    // in the chain
+    for (let i = path.length - 2; i >= 0; i--) {
+      if (!path[i].issuer.equals(path[i + 1].subject)) {
+        throw new CertificateChainVerificationError(
+          'Incorrect certificate name chaining'
```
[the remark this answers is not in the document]
bdehamer: Covering this case with a test turns out to be really difficult. The `openssl` tooling I'm using to generate my test certs won't let you do this. Creating test data for this case would require manually crafting/signing a certificate chain. I started down this path but realized the effort far outweighed the benefit of covering the negative condition here 😆

src/x509/verify.ts
bdehamer: Renamed

src/x509/verify.ts
feelepxyz: Ah yes this makes sense, seems ok to leave it in 👍

feelepxyz (approved):
🚢

PR #220, opened 2022-12-22
Serialize certificate as DER-encoded buffer

bdehamer:

#### Summary
Adds the plumbing necessary to re-serialize an x509 certificate back to a DER-encoded buffer after it has been parsed. This isn't used at the moment, but will be necessary when we start validating signed certificate timestamp (SCT) values which require that we remove the SCT extension and re-serialize the certificate for signature calculation.


feelepxyz (approved):
LGTM! 💯

PR #225, opened 2022-12-29
Fetch existing Rekor entry on create conflict

bdehamer:
#### Summary
Adds a `fetchOnConflict` option to `tlog.createDSSEEntry` which allows it to retrieve/return an existing Rekor entry in the event that the entry being creating matches an existing record. The default behavior is still to throw an error in the event of a conflict (HTTP status code 409), but the `fetchOnConflict` flag  can be enabled to override this behavior.

### DIFF
--- a/src/__tests__/tlog/index.test.ts
+++ b/src/__tests__/tlog/index.test.ts
@@ -204,7 +204,7 @@ describe('TLogClient', () => {
       toProposedIntotoEntry(dsse, sigMaterial)
     );
 
-    describe('when Rekor returns an error', () => {
+    describe('when Rekor returns a 500 error', () => {
       beforeEach(() => {
         nock(baseURL)
           .matchHeader('Accept', 'application/json')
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1541,23 +1541,6 @@
         "url": "https://opencollective.com/typescript-eslint"
       }
     },
-    "node_modules/@typescript-eslint/scope-manager": {
```
feelepxyz, line 1544: Do you know why there are a bunch of deletions here?

feelepxyz (dismissed):
Appreciate you implementing this! Looks good to me 👍  just had a question about the lockfile deletions, in case they where unintentional?

package-lock.json
bdehamer: I think this had something to do with dependabot. After the last batch of dependabot updates, I pulled down the main branch and ran `npm i` -- it resulted in all of these changes to the package-lock

bdehamer:
@feelepxyz today's round of dependabot updates seem to have cleaned-up the package-lock stuff. I rebased and everything is looking better now.

feelepxyz (approved):
🙌

PR #212, opened 2022-12-19
integrate bundle format changes

bdehamer:
#### Summary
Updates the generated bundle types to match the latest [changes](https://github.com/sigstore/protobuf-specs/pull/36/files) to the protobufs. This is a breaking change as the serialized bundle format is now different. The primary change is that the `tlogEntries` array is now part of the `verificationMaterial` object (the `verificationData` field has been removed from the bundle entirely).

Old Format:
```json
{
  "mediaType": "application/vnd.dev.sigstore.bundle+json;version=0.1",
  "verificationMaterial": {
    "x509CertificateChain": {
      "certificates": [
        {
[...]

PR #230, opened 2023-01-04
codegen new types from sigstore protobufs

bdehamer:

#### Summary
Rejiggering the code generation for the Sigstore protobufs. The Sigstore [protobufs](https://github.com/sigstore/protobuf-specs) now include more than just the definition for the bundle format, so putting the generated types into a directory named "bundle" no longer makes sense. The generated code is now placed in a directory called "sigstore".

Also, some of the new protobufs reference 3rd-party protos that were not present in the Docker image I was using to wrap the `protoc` CLI. I've switched to using the `namely/protoc-all` base image was already contains a number of commonly-referenced plugins and protobuf definitions. This is the same [base image](https://github.com/sigstore/protobuf-specs/blob/main/Dockerfile#L1) used by the Sigstore `protobuf-spec` project.


hack/images/protoc/Dockerfile
```diff
@@ -1,28 +1,5 @@
...
-###############
-FROM node:16.17
-
-WORKDIR /client
-
-COPY --from=protoc /protoc/bin/protoc /usr/local/bin
-
-ARG TSPROTO_VERSION=1.125.0
-RUN npm install -g "ts-proto@${TSPROTO_VERSION}"
+FROM namely/protoc-all@sha256:07f1ba9dbe11f5675e2efc8617c9552217dc4c3eb5ccd108f7c3889878dbae50
```
feelepxyz, line 2: Nice 👍

feelepxyz (approved):
Will this cause any issues because of the changed bundle format and not integrating these changes? https://github.com/sigstore/sigstore-js/pull/212

bdehamer:
> Will this cause any issues because of the changed bundle format and not integrating these changes? https://github.com/sigstore/sigstore-js/pull/212

For now these types will live alongside the older "bundle" types. I wanted to be able to use some of the verification-related types without being blocked by the breaking bundle changes.

After we merge #212 I'll fold all this together so that we have a single copy of the generated code.

PR #231, opened 2023-01-04
refactor ByteStream utility class

bdehamer:

#### Summary
Refactors the `ByteStream` utility class to support both reading-from and writing-to a stream. Since we need to both decode and encode x509 certificates its handy to have a data structure which allows us to either read-from or write-to an array of bytes in discrete chunks.


feelepxyz (approved):
LGTM 💯

PR #232, opened 2023-01-04
add createPublicKey to crypto utility module

bdehamer:

#### Summary
Adds a `createPublicKey` function to the crypto utility module which can be used to parse either PEM- or DER-encoded keys and turn them into `KeyLike` objects which can be passed to the sign and verify functions.

Also updates the `verifyBlob` function signature to accept an optional algorithm param which can be used to identify the hash algorithm which should be used when verifying the signature of an artifact.

### DIFF
--- a/src/util/crypto.ts
+++ b/src/util/crypto.ts
@@ -25,6 +25,14 @@ export function generateKeyPair(): KeyPairKeyObjectResult {
   });
 }
 
+export function createPublicKey(key: string | Buffer): KeyLike {
+  if (typeof key === 'string') {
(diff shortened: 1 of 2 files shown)


src/util/crypto.ts
```diff
@@ -35,12 +43,13 @@ export function signBlob(
 export function verifyBlob(
   data: Buffer,
   key: KeyLike,
-  signature: Buffer
+  signature: Buffer,
+  algorithm?: string
```
feelepxyz, line 47: Worth adding a test for this?

PR #236, opened 2023-01-05
verification of cert-embedded SCTs

bdehamer:

#### Summary
Adds a `verifySCTs` function to the `x509Certificate` class which will verify each of the signed-certificate timestamps in the SCT extension against the certificate's issuer and the transparency log public key.

### DIFF
--- a/src/x509/cert.ts
+++ b/src/x509/cert.ts
@@ -1,5 +1,6 @@
-import crypto from 'crypto';
-import { pem } from '../util';
+import * as sigstore from '../types/sigstore';
+import { crypto, pem } from '../util';
+import { ByteStream } from '../util/stream';
 import { ASN1Obj } from './asn1/obj';
 import {
   x509AuthorityKeyIDExtension,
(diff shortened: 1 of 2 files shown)


src/x509/cert.ts
```diff
@@ -157,14 +162,71 @@ export class x509Certificate {
...
+      sct.verify(preCert.buffer, logs)
+    );
+  }
+
+  // Creates a copy of the certificate with a new buffer
+  private clone(): x509Certificate {
+    const clone = Buffer.alloc(this.root.raw.length);
+    this.root.raw.copy(clone);
+    return x509Certificate.parse(clone);
+  }
```
bdehamer, line 224: Adding a `clone` method so that we can verify SCTs without mutating the original certificate (we need to be able to remove the SCT extension from the certificate and re-serialize it as part of the verification logic).

src/x509/cert.ts
```diff
@@ -157,14 +162,71 @@ export class x509Certificate {
...
+      }
+    }
 
+    if (!extSCT) {
+      throw new Error('Certificate does not contain SCT extension');
+    }
+
+    if (extSCT?.signedCertificateTimestamps?.length === 0) {
+      throw new Error('Certificate does not contain any SCTs');
+    }
```
bdehamer, line 199: Don't have a good way to cover this condition with a unit test. I don't have any example certificates which contain the SCT extension but have an empty list of SCTs.

src/x509/cert.ts
```diff
@@ -157,14 +162,71 @@ export class x509Certificate {
...
+    preCert.appendView(issuerId);
+
+    // Re-encodes the certificate to DER after removing the SCT extension
+    const tbs = clone.tbsCertificate.toDER();
+    preCert.appendUint24(tbs.length);
+    preCert.appendView(tbs);
+
+    return extSCT.signedCertificateTimestamps.every((sct) =>
+      sct.verify(preCert.buffer, logs)
+    );
```
bdehamer, line 216: When I integrate this into the top-level verification logic, I'll likely refactor this to return an array of results (or maybe the count of the number of verified SCTs) so that we can compare to the verification threshold.

feelepxyz (approved):
Looks great! Found the PR comments very helpful, great level of detail in comments made it easy to follow along 💯

PR #243, opened 2023-01-09
introduce new CA client class

bdehamer:

#### Summary
Introduces a new `CAClient` which sits between the `Signer` and the `Fulcio` client. This helps to encapsulate some of the details around formatting the Fulcio certificate request (which will come in handy when we switch to the v2 API at some point) and also mirrors the way that we have fronted the `Rekor` API client w/ a `TLog` class.

This will also give us a nice place to add the Fulcio-specific certificate verification logic.


feelepxyz (approved):
LGTM 👍

PR #245, opened 2023-01-09
bundle conversion helper functions

bdehamer:

#### Summary
Adds a couple of helper functions to convert from the legacy (current) bundle format to the v1 (proposed) format. These will be removed once we have migrated everything to the v1 format, but I don't want to write anymore code against the old format, so I'll use this shim as a temporary way to stitch together code relying on different versions of the bundle.

### DIFF
--- a/src/__tests__/types/index.test.ts
+++ b/src/__tests__/types/index.test.ts
@@ -1,0 +1,44 @@
+/*
+Copyright 2023 The Sigstore Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import * as legacy from '../../types/bundle';
+import {
+  translateLegacyBundle,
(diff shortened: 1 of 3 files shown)


src/__tests__/types/index.test.ts
```diff
@@ -0,0 +1,44 @@
...
+describe('translateLegacyBundle', () => {
+  it('should trantlates the legacy bundle', () => {
+    const legacyBundleJSON = bundles.dsse.valid.withSigningCert;
+    const legacyBundle = legacy.Bundle.fromJSON(legacyBundleJSON);
+    const bundle = translateLegacyBundle(legacyBundle);
+
+    expect(bundle).toBeDefined();
+    expect(bundle.verificationMaterial?.tlogEntries).toBeDefined();
+    expect(bundle.verificationMaterial?.tlogEntries).toHaveLength(1);
+  });
```
bdehamer, line 43: Not bothering w/ super-exhaustive testing here. This code will hopefully be short-lived and the TS compiler is doing most the heavy lifting here.

src/__tests__/types/index.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    const json = translateLegacyBundleJSON(legacyBundleJSON);
+
+    expect(json).toBeDefined();
+    expect(json.verificationMaterial.tlogEntries).toBeDefined();
+    expect(json.verificationData).toBeUndefined();
+  });
+});
+
+describe('translateLegacyBundle', () => {
+  it('should trantlates the legacy bundle', () => {
```
feelepxyz, line 35:
```suggestion
  it('should translate the legacy bundle', () => {
```

feelepxyz (approved):
Nice!

PR #248, opened 2023-01-10
verification of bundle certificates

bdehamer:

#### Summary
Adds a `verifySigningCertificate` function which accepts a Sigstore bundle and a trusted root and will perform all of the necessary verification steps on the certificate chain contained in the bundle (certificated chain verification, and verification of any SCTs in the signing certificate).

### DIFF
--- a/src/ca/verify.ts
+++ b/src/ca/verify.ts
@@ -1,0 +1,125 @@
+/*
+Copyright 2022 The Sigstore Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import { VerificationError } from '../error';
+import * as sigstore from '../types/sigstore';
+import { x509Certificate } from '../x509/cert';
+import { verifyCertificateChain } from '../x509/verify';
+
+export function verifySigningCertificate(
+  bundle: sigstore.Bundle,
+  trustedRoot: sigstore.TrustedRoot,
+  options: sigstore.ArtifactVerificationOptions_CtlogOptions
+) {
+  if (bundle.verificationMaterial?.content?.$case !== 'x509CertificateChain') {
+    throw new VerificationError('No certificate chain in bundle');
+  }
+
+  const bundleCerts = parseCerts(
+    bundle.verificationMaterial.content.x509CertificateChain.certificates
+  );
+
+  // Check that a trusted certificate chain can be found for the signing
+  // certificate in the bundle
+  const trustedChain = verifyChain(
+    bundleCerts,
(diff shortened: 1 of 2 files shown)


src/ca/verify.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  // Check that a trusted certificate chain can be found for the signing
+  // certificate in the bundle
+  const trustedChain = verifyChain(
+    bundleCerts,
+    trustedRoot.certificateAuthorities
+  );
+
+  // Unless disabled, verify the SCTs in the signing certificate
+  if (!options.disable) {
```
feelepxyz, line 42: Should this option be more explicit, like `disableSCTVerification`? Also when might we want to do this?

bdehamer: From the Sigstore Client Spec:
> If CT log verification is required for the signing certificate, verify that the certificate has a number of SCTs from the set of trusted CT log keys, larger or equal to the provided threshold.

But not really clear to me why CT log verification would **not** be required.

src/ca/verify.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+
+  if (validCAs.length === 0) {
+    throw new VerificationError('No valid certificate authorities');
+  }
+
+  let trustedChain: x509Certificate[] = [];
+
+  // Loop through all valid CAs and attempt to verify the certificate chain
+  const verified = validCAs.find((ca) => {
```
feelepxyz, line 67: Does `find` not return as soon as it you return true? Should this be a map?

src/ca/verify.ts
```diff
@@ -0,0 +1,125 @@
...
+  options: sigstore.ArtifactVerificationOptions_CtlogOptions
+) {
+  const signingCert = certificateChain[0];
+  const issuerCert = certificateChain[1];
+  const sctResults = signingCert.verifySCTs(issuerCert, ctLogs);
+
+  // Count the number of verified SCTs which were found
+  const verifiedSCTCount = sctResults.filter((sct) => sct.verified).length;
+
+  if (verifiedSCTCount < options.threshold) {
```
feelepxyz, line 100: Maybe ok, but wondering if this should be more explicit like `SCTVerificationThreshold `?

src/ca/verify.ts
feelepxyz: or wait, you only care that one valid CA has a valid cert chain? the comment tripped me up

src/ca/verify.ts
feelepxyz: Oh I see, yeah seems sound if this is specced out in the protobufs 👍 

@kommendorkapten do you have any context on why you you might want to disable CT log verification? Like for someone self-hosting Fulcio who hasn't set up SCTs?

src/ca/verify.ts
feelepxyz: Ref https://github.com/sigstore/sigstore-js/pull/248#discussion_r1066013245

feelepxyz (approved):
Think this looks good and seems good to align options to the client spec 🎉

src/ca/verify.ts
bdehamer: Agree that the naming is vague, but this is another case where we're using the types generated directly from the protobufs (https://github.com/sigstore/protobuf-specs/blob/a09e324bbc409628cef99a33e05fad2a5b4c47e5/protos/sigstore_verification.proto#L57-L66)

PR #249, opened 2023-01-11
narrow some of the sigstore generated types

bdehamer:

#### Summary
Establishes some "narrowed" versions of the types generated from the Sigstore protobufs. Specifically:

* `BundleWithCertificateChain` - a version of the Sigstore `Bundle` which is known to have a Fulcio certificate chain as part of the verification material.
* `CAArtifactVerificationOptions` a version of the `ArtifactVerificationOptions` which is known to contain the `ctlogOptions` and `signers` data necessary to verify a bundle containing a Fulcio certificate chain.

Using these narrowed types (and associated type guards) limits the number of places where we need to check for the presence of certain optional values as we pass these objects through the various layers of the system.

I expect to add more types like this as I build out the rest of the verification logic.

### DIFF
--- a/src/__tests__/ca/verify.test.ts
+++ b/src/__tests__/ca/verify.test.ts
@@ -21,7 +21,9 @@ import bundles from '../__fixtures__/bundles/';
 describe('verifySigningCertificate', () => {
   // Temporary until we reconsole bundle formats
   const bundleJSON = bundles.dsse.valid.withSigningCert;
-  const bundle = sigstore.Bundle.fromJSON(bundleJSON);
+  const bundle = sigstore.Bundle.fromJSON(
+    bundleJSON
+  ) as sigstore.BundleWithCertificateChain;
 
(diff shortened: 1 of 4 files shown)


src/__tests__/ca/verify.test.ts
```diff
@@ -24,7 +24,9 @@ describe('verifySigningCertificate', () => {
   const bundleJSON = translateLegacyBundleJSON(
     bundles.dsse.valid.withSigningCert
   );
-  const bundle = sigstore.Bundle.fromJSON(bundleJSON);
+  const bundle = sigstore.Bundle.fromJSON(
+    bundleJSON
+  ) as sigstore.BundleWithCertificateChain;
```
bdehamer, line 29: Since this is just test data and I know exactly what I've loaded, I'm just going to cast it to the type I need for the tests.

feelepxyz (dismissed):
Nice refactor 👍

PR #252, opened 2023-01-11
combine generated types

bdehamer:

#### Summary
This is the final clean-up of the bundle migration effort. During the transition period we had two versions of the code-generated Sigstore types (one in "src/types/bundle" and another in "src/types/sigstore"). Now that we have completed the transition to the newer bundle format (and no longer need to support the old format) this PR cleans-up the dead code and updates all references to point to the new types.


feelepxyz (approved):
🎉

PR #253, opened 2023-01-12
signer identity verification

bdehamer:
#### Summary

Adds a new `verifySignerIdentity` function which will verify the Fulcio-issued signing certificate against the signer identity contained in the `ArtifactVerificationOptions`
```
verifySignerIdentity
  when there is a matching identity
    when only one identity is specified
      ✓ returns without error
    when the matching identity has an empty OID list
      ✓ returns without error
    when the identity list contains a non-matching identity
      ✓ returns without error
  when there is NO matching identity
    when the issuer does not match
      ✓ throws an error
    when the certificate does NOT have a SAN extension
      ✓ throws an error
    when the required SAN is NOT specified
      ✓ throws an error
    when the required SAN identity is NOT specified
      ✓ throws an error 
    when the required SAN type is NOT specified
      ✓ throws an error
    when the required SAN type is OTHER_NAME (not supported yet)
      ✓ throws an error
    when the required SAN type does NOT match the certificate
      ✓ throws an error 
    when the SAN value is a regexp (not supported yet)
      ✓ throws an error
    when a required extension is missing
      ✓ throws an error
    when a required extension does NOT specify an OID
      ✓ throws an error
```

### DIFF
--- a/hack/ca/Makefile
+++ b/hack/ca/Makefile
@@ -97,6 +97,7 @@ nosan.crt: int.crt
 		-subj / \
 		-addext "keyUsage = critical, digitalSignature" \
 		-addext "extendedKeyUsage = codeSigning" \
+		-addext "1.3.6.1.4.1.57264.1.1 = DER:46:4F:4F" \
 		-out store/nosan/nosan.csr
 	openssl ca \
 		-notext \
--- a/src/__tests__/ca/verify/index.test.ts
+++ b/src/__tests__/ca/verify/index.test.ts
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Sigstore Authors.
+Copyright 2023 The Sigstore Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -42,7 +42,21 @@ describe('verifySigningCertificate', () => {
   const signers: sigstore.CAArtifactVerificationOptions['signers'] = {
     $case: 'certificateIdentities',
     certificateIdentities: {
-      identities: [],
+      identities: [
+        {
+          issuer: 'https://github.com/login/oauth',
+          san: {
+            type: sigstore.SubjectAlternativeNameType.EMAIL,
+            identity: {
+              $case: 'value',
+              value: Buffer.from('YnJpYW5AZGVoYW1lci5jb20=', 'base64').toString(
+                'utf-8'
+              ),
+            },
+          },
+          oids: [],
+        },
+      ],
     },
   };
 
--- a/src/__tests__/ca/verify/signer.test.ts
+++ b/src/__tests__/ca/verify/signer.test.ts
@@ -1,0 +1,322 @@
+/*
+Copyright 2023 The Sigstore Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import { verifySignerIdentity } from '../../../ca/verify/signer';
+import * as sigstore from '../../../types/sigstore';
+import { x509Certificate } from '../../../x509/cert';
+import { certificates } from '../../__fixtures__/certs';
+
+describe('verifySignerIdentity', () => {
+  const signingCert = x509Certificate.parse(certificates.fulcioleaf);
+
+  const issuer = 'https://token.actions.githubusercontent.com';
+  const workflowURI =
+    'https://github.com/sigstore/sigstore-js/.github/workflows/publish.yml@refs/tags/v0.4.0';
+
+  const workflowSAN: sigstore.SubjectAlternativeName = {
+    type: sigstore.SubjectAlternativeNameType.URI,
+    identity: {
+      $case: 'value',
+      value: workflowURI,
+    },
+  };
+
+  const extFulcioWorkflowName: sigstore.ObjectIdentifierValuePair = {
+    oid: { id: [1, 3, 6, 1, 4, 1, 57264, 1, 4] },
+    value: Buffer.from('publish'),
+  };
+
+  const extFulcioRepository: sigstore.ObjectIdentifierValuePair = {
+    oid: { id: [1, 3, 6, 1, 4, 1, 57264, 1, 5] },
+    value: Buffer.from('sigstore/sigstore-js'),
+  };
+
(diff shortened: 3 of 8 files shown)


hack/ca/Makefile
```diff
@@ -97,6 +97,7 @@ nosan.crt: int.crt
 		-subj / \
 		-addext "keyUsage = critical, digitalSignature" \
 		-addext "extendedKeyUsage = codeSigning" \
+		-addext "1.3.6.1.4.1.57264.1.1 = DER:46:4F:4F" \
```
bdehamer, line 100: Updating my certificate test fixtures to include a cert which has a Fulcio issuer extension, but not SAN extension.

src/ca/verify/index.ts
```diff
@@ -33,4 +34,6 @@ export function verifySigningCertificate(
   if (options.ctlogOptions.disable === false) {
     verifySCTs(trustedChain, trustedRoot.ctlogs, options.ctlogOptions);
   }
+
+  verifySignerIdentity(trustedChain[0], options.signers.certificateIdentities);
```
bdehamer, line 38: Verification of the signing certificate will now include:

* Verification that the certificate chains back to a trusted root certificate
* Verification of signed certificate timestamps (SCTs) embedded in the certificate
* Verification that the issuer/SAN of the certificate matches some expected value

src/ca/verify/signer.ts
```diff
@@ -0,0 +1,136 @@
...
+  let sanValue: string | undefined;
+  switch (expectedSAN.type) {
+    case sigstore.SubjectAlternativeNameType.EMAIL:
+      sanValue = sanExtension.rfc822Name;
+      break;
+    case sigstore.SubjectAlternativeNameType.URI:
+      sanValue = sanExtension.uri;
+      break;
+    case sigstore.SubjectAlternativeNameType.OTHER_NAME:
+    // TODO: Support other name types
```
bdehamer, line 98: Support for this is going to require some work in the `x509SubjectAlternativeNameExtension` class in order to retrieve the `OtherName` value (see [rfc5280](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.6)). Gonna tackle this in a separate PR.

src/ca/verify/signer.ts
```diff
@@ -0,0 +1,136 @@
...
+    return false;
+  }
+
+  let match: string | undefined;
+  switch (expectedSAN.identity.$case) {
+    case 'value':
+      match = expectedSAN.identity.value;
+      break;
+    case 'regexp':
+      // TODO support regex
```
bdehamer, line 112: I have lots of questions about how this is supposed to work. Gonna circle back on supporting regex matches for SAN values.

src/x509/cert.ts
```diff
@@ -146,6 +161,11 @@ export class x509Certificate {
     return ca;
   }
 
+  public extension(oid: string): x509Extension | undefined {
+    const ext = this.findExtension(oid);
+    return ext ? new x509Extension(ext) : undefined;
+  }
+
```
bdehamer, line 168: Updating the `x509Certificate` class to support the retrieval of arbitrary extensions by OID in support of the OID verification checks (i.e. we need to be able to check that a certificate contains specific extensions with specific values).

src/__tests__/ca/verify/index.test.ts
```diff
@@ -42,7 +42,21 @@ describe('verifySigningCertificate', () => {
   const signers: sigstore.CAArtifactVerificationOptions['signers'] = {
     $case: 'certificateIdentities',
     certificateIdentities: {
-      identities: [],
+      identities: [
+        {
+          issuer: 'https://github.com/login/oauth',
```
feelepxyz, line 47: This is nice! For registry, will it be possible to check issuer and oiods without checking for a particular `san`? (to validate part of the provenance contents against the cert)

src/ca/verify/signer.ts
```diff
@@ -0,0 +1,136 @@
...
+import * as sigstore from '../../types/sigstore';
+import { x509Certificate } from '../../x509/cert';
+
+// https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#1361415726411--issuer
+const OID_FUCIO_ISSUER = '1.3.6.1.4.1.57264.1.1';
+
+// Verifies the identity embedded in a Fulcio-issued signing certificate against
+// the list of trusted identities. Returns without error if at least one of the
+// identities matches the signing certificate; otherwise, throws a
+// VerificationError.
```
feelepxyz, line 26: This is a helpful comment!

src/__tests__/ca/verify/index.test.ts
feelepxyz: Ah I see this is required now, we can probably figure out what it should be during publish.

src/ca/verify/signer.ts
feelepxyz: Good call! Seems like it should really be a strict subset of regex, like only allow simple globs?

feelepxyz: Love that we're implementing this so close to the specs and raising these questions.

src/x509/cert.ts
```diff
@@ -146,6 +161,11 @@ export class x509Certificate {
     return ca;
   }
 
+  public extension(oid: string): x509Extension | undefined {
```
feelepxyz, line 164: nit pick rename? somehow this name makes me think it's a setter
```suggestion
  public getExtension(oid: string): x509Extension | undefined {
```

src/__tests__/ca/verify/signer.test.ts
```diff
@@ -0,0 +1,322 @@
...
+  const workflowSAN: sigstore.SubjectAlternativeName = {
+    type: sigstore.SubjectAlternativeNameType.URI,
+    identity: {
+      $case: 'value',
+      value: workflowURI,
+    },
+  };
+
+  const extFulcioWorkflowName: sigstore.ObjectIdentifierValuePair = {
+    oid: { id: [1, 3, 6, 1, 4, 1, 57264, 1, 4] },
```
feelepxyz, line 37: Any reason this isn't a string? Is this set in the spec?

src/x509/cert.ts
feelepxyz: I see we have `extensions()` already, also ok as is 👍

feelepxyz (approved):
This looks great! Appreciate the countless bite sized PRs you've been feeding me, which has made reviewing easy!

src/__tests__/ca/verify/signer.test.ts
bdehamer: @feelepxyz yeah, I thought this was odd as well, but it's how it's defined in the protobufs:

```protobuf
message ObjectIdentifier {
        repeated int32 id = 1 [(google.api.field_behavior) = REQUIRED];
}
```

I was digging through some Go code yesterday and saw this:
```golang
var oidpSpecified = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 1, 9}
```
So I'm guess there is some established convention ¯\_(ツ)_/¯

PR #254, opened 2023-01-12
SAN other name

bdehamer:
#### Summary
Updates the `x509SubjectAlternativeNameExtension` class to support the retrieval of "otherName"-style SAN values. Fulcio may use "otherName" SAN values w/ a custom OID and we need to be able to read those values during certificate verification.


hack/ca/Makefile
```diff
@@ -56,7 +56,7 @@ leaf.crt: int.crt
 		-subj / \
 		-addext "keyUsage = critical, digitalSignature" \
 		-addext "extendedKeyUsage = codeSigning" \
-		-addext "subjectAltName = URI:http://foobar.dev" \
+		-addext "subjectAltName = URI:http://foobar.dev,otherName:1.3.6.1.4.1.57264.1.7;UTF8:FOO" \
```
bdehamer, line 59: Sets-up a test certificate with two SANs, one with the "URI" type and one with the "otherName" type:

```
SEQUENCE (2 elem)
  OBJECT IDENTIFIER 2.5.29.17 subjectAltName (X.509 extension)
    OCTET STRING (42 byte) 
      SEQUENCE (2 elem)
        [6] (17 byte) http://foobar.dev
        [0] (2 elem)
          OBJECT IDENTIFIER 1.3.6.1.4.1.57264.1.7
          [0] (1 elem)
            UTF8String FOO
```

src/x509/asn1/tag.ts
```diff
@@ -51,7 +51,7 @@ export class ASN1Tag {
 
   public isContextSpecific(num?: number): boolean {
     const res = this.class === TAG_CLASS.CONTEXT_SPECIFIC;
-    return num ? res && this.number === num : res;
+    return num !== undefined ? res && this.number === num : res;
```
bdehamer, line 54: Bug fix in `ASN1Tag.isContextSpecific`. Wasn't distinguishing between a missing `num` param and a falsy value like `0x00`

feelepxyz (dismissed):
🙌

PR #256, opened 2023-01-13
revisit implementation for verifyTLogSET

bdehamer:

#### Summary
Re-implementation of the `verifyTLogSET` function to account for the fact that we need to be able check the SET against multiple trusted TLog instances -- returning `true` as long as we get verification against one of them.

```
verifyTLogSET
  when there is a matching TLogInstance
    when the SET can be verified
      ✓ returns true 
    when the SET can NOT be verified
      ✓ returns false
    when the matching TLogInstance has no public key
      ✓ returns false
    when the matching TLogInstance has an empty public key
      ✓ returns false
    when the public key for the matching TLogInstance is not valid
      when public key has no start
        ✓ returns false
      when the public key has a start after the integrated time
        ✓ returns false
      when the public key has an end before the integrated time
        ✓ returns false
  when there is NO matching TLogInstance
    ✓ returns false
```

### DIFF
--- a/src/tlog/verify/set.ts
+++ b/src/tlog/verify/set.ts
@@ -1,0 +1,117 @@
+/*
+Copyright 2023 The Sigstore Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+import * as sigstore from '../../types/sigstore';
+import { crypto, json } from '../../util';
+
+// Structure over which the tlog SET signature is generated
+interface VerificationPayload {
+  body: string;
+  integratedTime: number;
+  logIndex: number;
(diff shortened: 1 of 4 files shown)


src/tlog/verify/set.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Verifies the SET for the given entry against the list of trusted
+// transparency logs. Returns true if the SET can be verified against at least
+// one of the trusted logs; otherwise, returns false.
+export function verifyTLogSET(
+  entry: sigstore.VerifiableTransparencyLogEntry,
+  tlogs: sigstore.TransparencyLogInstance[]
+): boolean {
+  // Filter the list of tlog instances to only those which which match the given
```
feelepxyz, line 34:
```suggestion
  // Filter the list of tlog instances to only those which match the given
```

src/tlog/verify/set.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Filter the list of tlog instances to only those which which match the given
+// log ID and have public keys which are valid for the given integrated time.
+function filterTLogInstances(
+  tlogInstances: sigstore.TransparencyLogInstance[],
+  logID: Buffer,
+  integratedTime: string
+): sigstore.TransparencyLogInstance[] {
+  // Convert the integrated time to a date object for comparison
```
feelepxyz, line 86: nit: not sure this comment adds much beyond the code?
```suggestion
```

feelepxyz (approved):
Minor comments otherwise LGTM 👍

PR #257, opened 2023-01-14
revisit implementation for verifyTLogBody

bdehamer:

#### Summary
Re-implementation of the `verifyTLogBody` function to account for the fact that we need to be able check multiple tlog entries and ensure that some minimum threshold of them check out.

```
verifyTLogBody
  when a message signature bundle is provided
    when everything is valid
      ✓ returns true 
    when the signature does NOT match the value in the tlog entry
      ✓ returns false
    when the digest does NOT match the value in the tlog entry
      ✓ returns false
    when there is a version mismatch between the tlog entry and the body
      ✓ returns false
  when a DSSE Bundle is provided
    when everything is valid
      ✓ returns false
    when the payload hash does NOT match the value in the intoto entry
      ✓ returns false
    when the signature does NOT match the value in the intoto entry
      ✓ returns false
    when the tlog entry version is unsupported
      ✓ returns false
    when the signature count does NOT match the intoto entry
      ✓ returns false
    when there is a version mismatch between the tlog entry and the body
      ✓ returns false

```


src/tlog/verify/body.ts
```diff
@@ -0,0 +1,159 @@
...
+  dsse: sigstore.Envelope
+): void {
+  // Collect all of the signatures from the DSSE envelope
+  // Turns them into base64-encoded strings for comparison
+  const dsseSigs = dsse.signatures.map((signature) =>
+    signature.sig.toString('base64')
+  );
+
+  // Collect all of the signatures from the tlog entry
+  // Remember that tlog signastures are double base64-encoded
```
feelepxyz, line 114: 🤦

src/__tests__/tlog/verify/body.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  describe('when a DSSE Bundle is provided', () => {
+    describe('when everything is valid', () => {
+      const bundle = sigstore.bundleFromJSON(
+        bundles.dsse.valid.withSigningCert
+      );
+      const tlogEntry = bundle.verificationMaterial
+        ?.tlogEntries[0] as sigstore.VerifiableTransparencyLogEntry;
+
+      it('returns false', () => {
```
feelepxyz, line 80:
```suggestion
      it('returns true', () => {
```

src/tlog/verify/body.ts
```diff
@@ -0,0 +1,159 @@
...
+limitations under the License.
+*/
+import { UnsupportedVersionError, VerificationError } from '../../error';
+import * as sigstore from '../../types/sigstore';
+import { crypto, encoding as enc } from '../../util';
+import { EntryKind, HashedRekordKind, IntotoKind } from '../types';
+
+const TLOG_MISMATCH_ERROR_MSG = 'bundle content and tlog entry do not match';
+
+// Compare the given intoto tlog entry to the given bundle
```
feelepxyz, line 23: Should this say intoto as it also checks hashedrekord/mesasgetype? 
```suggestion
// Compare the given tlog entry to the given bundle
```

feelepxyz (approved):
Look good! Few minor comments.

PR #262, opened 2023-01-17
tools: automate GitHub release creation

MylesBorins:
This extends the `publish.yml` workflow to automate the creation of a GitHub release with automated release notes

Signed-off-by: Myles Borins <mylesborins@github.com>

#### Summary

This further automates the release process publishing to npm + creating a GitHub release from any pushed tag

#### Release Note

* update `publish.yml` to automate GitHub releases

#### Documentation

N/A

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -27,3 +27,17 @@ jobs:
       run: npm exec npm@npm/cli#provenance -- publish --provenance
       env:
         NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
+
+  create-github-release:
+    name: Create GitHub Release
+    runs-on: ubuntu-latest
+    needs: publish
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3
+      - name: Create Release
+        run: gh release create ${{ github.ref }} --generate-notes
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,8 +1,9 @@
 name: publish
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v*'
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -1,8 +1,9 @@
 name: publish
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v*'
```
MylesBorins, line 6: With this change you can locally do

```
$ npm version patch|minor|major
$ git push origin main --follow-tags
```

and GitHub will do the rest 👍🏼

MylesBorins (commented):
i also have a manually triggered workflow that can be used to create the tag, can't propose that in a follow up PR

.github/workflows/publish.yml
bdehamer: @MylesBorins can you update the publish section in the readme with this bit

bdehamer (approved):
I love this! Thanks so much @MylesBorins 

I haven't used the `gh release` command before. Any tips on how we should format our commit messages so that we get something useful/readable out of this?

feelepxyz:
Nice! 

> Any tips on how we should format our commit messages so that we get something useful/readable out of this?

@bdehamer looks like this can be configured somewhat using a config file: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configurations

I'm still curious if we should adopt something like [changesets](https://github.com/changesets/changesets) to generate the changelogs from structured files to you can introduce with each PR, automatically running `npm version x` for you etc. It also comes with an action to create a "release" PR with all changes for the next release rolled up, based on the changesets you've created between the last release and default branch head.

Saying that, changesets feels somewhat heavy though and I like the simplify of relying on github building blocks as much as possible.

MylesBorins:
Looking at the current release notes it looks like you are already using the automatic release note generation feature from GitHub releases, the GH CLI uses the same feature.

It uses the title of the PR not individual commits

.github/workflows/publish.yml
MylesBorins: done PTAL

feelepxyz:
Closing this as we're now using changesets which creates the github release for us amongst other things: https://github.com/sigstore/sigstore-js/pull/339

PR #263, opened 2023-01-17
verify tlog entries from bundle

bdehamer:
#### Summary
Ties together the work in https://github.com/sigstore/sigstore-js/pull/256 and https://github.com/sigstore/sigstore-js/pull/257. The primary difference here from the original tlog verification implementation is that we now attempt to verify all the entries in the bundle and then check the count of verified entries against the configured threshold.


feelepxyz (approved):
LGTM!

PR #268, opened 2023-01-20
Check bundle for required fields

bdehamer:
#### Summary
Introduces a new narrowed version of the `Bundle` type named `ValidBundle` where all of the required fields are mandatory. Ideally this would be code-generated automatically from the [protobufs](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto), but the TypeScript code generator doesn't recognize the `field_behavior=REQUIRED` annotations (already opened as an [issue](https://github.com/stephenh/ts-proto/issues/164)).

The new type is paired with an `assertValidBundle` type assertion helper which will perform the necessary checks (throwing an `InvalidBundleError` if necessary) and automatically narrow the `Bundle` to a `ValidBundle`.

```
assertValidBundle
  when the bundle is completely empty
    ✓ throws an error 
  when the content type is messageSignature
    when the messageDigest is undefined
      ✓ throws an error
    when the digest value is undefined
      ✓ throws an error 
    when the signature value is undefined
      ✓ throws an error
  when the content type is dsseEnvelope
    when the payload is undefined
      ✓ throws an error
    when there are too few signatures
      ✓ throws an error 
    when there are too many signatures
      ✓ throws an error
    when there the signature value is undefined
      ✓ throws an error
  when the verificationMaterial is undefined
    ✓ throws an error
  when the verificationMaterial content is undefined
    ✓ throws an error
  when the verificationMaterial content type is x509CertificateChain
    when the certificate chain is empty
      ✓ throws an error 
    when the certificate chain is missing a certificate
      ✓ throws an error
  when everything is valid
    ✓ does NOT throw an error
```
In response to: https://sigstore.slack.com/archives/C03SZ6SHU90/p1674241965238669?thread_ts=1674238410.807909&cid=C03SZ6SHU90


src/types/sigstore/validate.ts
```diff
@@ -0,0 +1,88 @@
...
+export function assertValidBundle(b: Bundle): asserts b is ValidBundle {
+  // Content-related validation
+  if (b.content === undefined) {
+    throw new InvalidBundleError('content is undefined');
+  }
+
+  switch (b.content.$case) {
+    case 'messageSignature':
+      if (b.content.messageSignature.messageDigest === undefined) {
+        throw new InvalidBundleError('messageDigest is undefined');
```
vlsi, line 31: Is there a way you could use full property paths (e.g. messageSignature.messageDigest) in the error message?
Is there a way you could mention all the missing fields rather than stopping on the first missing one?

bdehamer: Good suggestions. Done!

src/types/sigstore/validate.ts
```diff
@@ -0,0 +1,97 @@
...
+  const invalidValues: string[] = [];
+
+  // Content-related validation
+  if (b.content === undefined) {
+    invalidValues.push('content');
+  } else {
+    switch (b.content.$case) {
+      case 'messageSignature':
+        if (b.content.messageSignature.messageDigest === undefined) {
+          invalidValues.push('content.messageSignature.messageDigest');
```
feelepxyz, line 32: Nice! 👍

feelepxyz (dismissed):
👍

feelepxyz (approved):

PR #273, opened 2023-01-24
integrate ValidBundle type

bdehamer:

#### Summary
Updates the interface of the `Verifier` to use the new `ValidBundle` type (which narrows the standard `Bundle` to a type for which we know all of the required fields have been populated). This allows us to remove a bunch of if-defined checks we have scattered across the code (and the associated unit tests).

### DIFF
--- a/src/types/sigstore/index.ts
+++ b/src/types/sigstore/index.ts
@@ -18,6 +18,7 @@ import { encoding as enc, pem } from '../../util';
 import { x509Certificate } from '../../x509/cert';
 import { SignatureMaterial } from '../signature';
 import { WithRequired } from '../utility';
+import { assertValidBundle, ValidBundle } from './validate';
 import { Envelope } from './__generated__/envelope';
 import { Bundle, VerificationMaterial } from './__generated__/sigstore_bundle';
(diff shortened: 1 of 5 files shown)


src/types/sigstore/index.ts
```diff
@@ -34,7 +35,14 @@ export * from './__generated__/sigstore_trustroot';
...
 
 export const bundleToJSON = Bundle.toJSON;
-export const bundleFromJSON = Bundle.fromJSON;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const bundleFromJSON = (obj: any): ValidBundle => {
+  const bundle = Bundle.fromJSON(obj);
+  assertValidBundle(bundle);
+  return bundle;
+};
```
bdehamer, line 44: Adding the call to `assertValidBundle` here means that we'll automatically validate the bundle at the point that we deserialize it from JSON. If this function returns, it is guaranteed to return a `ValidBundle`.

feelepxyz: Nice! Types FTW 🙌

feelepxyz (approved):
Awesome refactor!

PR #274, opened 2023-01-27
TUF integration

bdehamer:
#### Summary
Integrates the [`tuf-js`](https://github.com/theupdateframework/tuf-js) library and wires it up to the Sigstore TUF repository.

The current implementation queries the TUF repo for the various Rekor/Fulcio/CTFE keys/certs and assembles them into a structure which matches `TrustedRoot` [protobuf](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L70-L88). At some point, the Sigstore TUF repo will provide a serialized version of the `TrustedRoot` directly and a lot of this code will go away (https://github.com/sigstore/root-signing/pull/584).


src/tuf/index.ts
```diff
@@ -1,9 +1,80 @@
...
+  cacheDir: string
+): Promise<sigstore.TrustedRoot> {
+  initTufCache(cacheDir);
+  const repoMap = initRepoMap(cacheDir);
+
+  const repoClients = Object.entries(repoMap.repositories).map(([name, urls]) =>
+    initClient(name, urls[0], cacheDir)
+  );
+
+  // TODO: Add support for multiple repositories. For now, we just use the first
```
feelepxyz, line 27: Seems good to begin with 👍

src/tuf/index.ts
```diff
@@ -1,9 +1,80 @@
...
+    const tufRootSrc = require.resolve(`../../store/${name}-root.json`);
+    fs.mkdirSync(repoCachePath);
+    fs.copyFileSync(tufRootSrc, tufRootDest);
+  }
+
+  if (!fs.existsSync(targetPath)) {
+    fs.mkdirSync(targetPath);
+  }
+
+  // TODO: Is there some better way to derive the base URL for the targets?
```
feelepxyz, line 72: Sgtm for now 👍

feelepxyz (commented):
Awesome! Looks broadly good, the one thing I'm wondering about is error handling in downstream users like npm. Will we be able to catch and show helpful errors or retry? Worth adding any tests around errors fetching these tuf files?

src/sigstore.ts
```diff
@@ -111,7 +113,8 @@ export async function verify(
   options: VerifyOptions,
   data?: Buffer
 ): Promise<void> {
-  const trustedRoot = tuf.getTrustedRoot();
+  const cacheDir = path.join(os.homedir(), '.sigstore', 'cache');
```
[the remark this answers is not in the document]
bdehamer: Added a check to ensure that homedir is readable/writable with a fallback to the system tempdir. These are both functions from the node standard library which work across all platforms.

bdehamer:
> I'm wondering about is error handling in downstream users like npm. Will we be able to catch and show helpful errors or retry? Worth adding any tests around errors fetching these tuf files?

I updated things a bit so that errors thrown from our TUF client are caught and rethrown as something that will be more friendly to `sigstore-js` consumers. Retry on error is already something baked-in to the TUF client so we don't need to do anything here.

src/__tests__/tuf/trustroot.test.ts
```diff
@@ -0,0 +1,134 @@
...
+        const root = await subject.getTrustedRoot();
+
+        expect(root).toBeDefined();
+        expect(root.certificateAuthorities).toHaveLength(1);
+        expect(root.tlogs).toHaveLength(1);
+        expect(root.ctlogs).toHaveLength(1);
+        expect(root.timestampAuthorities).toHaveLength(0);
+
+        const ca = root.certificateAuthorities[0];
+        expect(ca.uri).toEqual('');
```
kommendorkapten, line 78: Might be misunderstanding something here, but I would expect the `uri` to be `https://fulcio.sigstore.dev`?

src/tuf/index.ts
```diff
@@ -1,9 +1,80 @@
...
+  const buf = fs.readFileSync(mapDest);
+  return JSON.parse(buf.toString('utf-8'));
+}
+
+function initClient(name: string, url: string, rootDir: string): Updater {
+  const repoCachePath = path.join(rootDir, name);
+  const targetCachePath = path.join(rootDir, 'targets');
+  const tufRootDest = path.join(repoCachePath, 'root.json');
+
+  // Only copy the trusted root if it doesn't already exist. It's possible that
```
kommendorkapten, line 59:
```suggestion
  // Only copy the TUF root if it doesn't already exist. It's possible that
```

kommendorkapten: Or maybe trusted TUF root, to avoid possible confusion with the "trusted root" which is a target in the TUF repository.

src/tuf/trustroot.ts
```diff
@@ -0,0 +1,159 @@
...
+        // The log ID is not present in the Sigstore target metadata, but
+        // can be derived by hashing the contents of the public key.
+        return {
+          baseUrl: uri,
+          hashAlgorithm: sigstore.HashAlgorithm.SHA2_256,
+          logId: { keyId: crypto.hash(keyBytes) },
+
+          publicKey: {
+            keyDetails: sigstore.PublicKeyDetails.PKIX_ECDSA_P256_SHA_256,
+            rawBytes: keyBytes,
```
kommendorkapten, line 96: This is not really right? The Key on disk is PEM encoded but in the bundle we only expect the base64 encoded DER bytes, i.e the PEM headers should be stripped.

kommendorkapten: Sorry, missed that `readKey` extracts the correct bytes.

src/tuf/trustroot.ts
```diff
@@ -0,0 +1,159 @@
...
+        certChain: {
+          certificates: certs.map((cert) => ({ rawBytes: cert })),
+        },
+      },
+    ];
+  }
+
+  // Reads the contents of the specified target file, and returns the
+  // contents as a DER-encoded buffer.
+  private async readKey(target: TargetFile): Promise<Buffer> {
```
kommendorkapten, line 136: Maybe change the name of this to `pemToDER` or something similar as it's what it's doing. It's not limited to a key.

src/tuf/trustroot.ts
```diff
@@ -0,0 +1,159 @@
...
+      // An empty path here means the target has not been cached locally, or is
+      // out of date. In either case, we need to download it.
+      if (!path) {
+        path = await this.tuf.downloadTarget(target);
+      }
+
+      const file = fs.readFileSync(path);
+      return pem.toDER(file.toString('utf-8'));
+    } catch (err) {
+      throw new TrustedRootError('error reading key');
```
kommendorkapten, line 149: Same here, this message could potentially be confusing when reading a certificate (which does contain a public key though 😄 )

src/sigstore.ts
```diff
@@ -149,6 +153,17 @@ function configureIdentityProviders(
...
 
+function defaultCacheDir(): string {
+  let cacheRootDir = os.homedir();
+  try {
+    fs.accessSync(os.homedir(), fs.constants.W_OK | fs.constants.R_OK);
+  } catch (e) {
+    cacheRootDir = os.tmpdir();
+  }
+
+  return path.join(cacheRootDir, '.sigstore', 'cache');
```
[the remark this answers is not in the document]
kommendorkapten: Coming back to this, I think the default cache dir should be:
`~/.sigstore/root/<name>` where `<name>` is the name in `map.json`, i.e `public-good-instance`. 
`map.json` should be stored in `~/.sigstore/root`. Targets can be stored in `~/.sigstore/root/<name>/targets`.

kommendorkapten:
Nice job 👍

src/sigstore.ts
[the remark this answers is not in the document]
bdehamer: @kommendorkapten here's what my local `~/.sigstore` looks like after updating with your recommendations (showing both `cosign` and `sigstore-js` TUF caches):
```
/Users/bdehamer/.sigstore
├── js-root
│   ├── map.json
│   └── public-good-instance
│       ├── root.json
│       ├── snapshot.json
│       ├── targets
│       │   ├── ctfe.pub
│       │   ├── ctfe_2022.pub
│       │   ├── fulcio.crt.pem
│       │   ├── fulcio_intermediate_v1.crt.pem
│       │   ├── fulcio_v1.crt.pem
│       │   └── rekor.pub
│       ├── targets.json
│       └── timestamp.json
└── root
    ├── remote.json
    ├── targets
    │   ├── artifact.pub
    │   ├── ctfe.pub
    │   ├── ctfe_2022.pub
    │   ├── fulcio.crt.pem
    │   ├── fulcio_intermediate_v1.crt.pem
    │   ├── fulcio_v1.crt.pem
    │   └── rekor.pub
    └── tuf.db
        ├── 000002.log
        ├── CURRENT
        ├── CURRENT.bak
        ├── LOCK
        ├── LOG
        └── MANIFEST-000003
```

How does that look?

src/__tests__/tuf/trustroot.test.ts
bdehamer: Since I'm constructing a single certificate authority entry from multiple 'Fulcio' targets, I didn't bother to collect/combine the URIs to populate this field (also, it's not being used anywhere else in the system at the moment).

This will all be much cleaner once we start consuming the actual `TrustedRoot`.

src/tuf/trustroot.ts
bdehamer: I find myself frustrated that we don't have a single word that encompasses "key or certificate". I updated this function to be named `readTargetBytes` (avoiding any specifics about WHAT I'm reading, but also making it clear that you're getting bytes in the response)

src/tuf/trustroot.ts
bdehamer: updated

src/sigstore.ts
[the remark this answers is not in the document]
kommendorkapten: I think this looks good. In the future we probably want to share the root (i.e only a single `map.json` for all clients), but now I think this is great to keep them disjoint 👍

kommendorkapten (approved):
Nice work!

src/sigstore.ts
```diff
@@ -149,6 +154,17 @@ function configureIdentityProviders(
   return idps;
 }
 
+function defaultCacheDir(): string {
+  let cacheRootDir = os.homedir();
+  try {
+    fs.accessSync(os.homedir(), fs.constants.W_OK | fs.constants.R_OK);
```
feelepxyz, line 160: TIL about `fs.access`, nice!

feelepxyz:
Awesome work getting this in! 🥳

PR #296, opened 2023-02-08
expose configurable TrustedRoot resolver

bdehamer:
#### Summary
Exposes a configurable `trustedRootResolver` as part of the verification options to help in testing situations where we may want to call `verify` but not interact w/ the Sigstore TUF repository.


bdehamer:
Decided against this approach. Client who don't want to interact w/ TUF will just have to stub their interaction with `sigstore-js`

PR #309, opened 2023-02-21
Strict ts-config

feelepxyz:
Added all the strict options I could find to the ts-config.


src/cli/index.ts
```diff
@@ -92,7 +92,7 @@ async function verify(bundlePath: string, artifactPath: string) {
   }
 
   const bundleFile = fs.readFileSync(bundlePath);
-  const bundle = JSON.parse(bundleFile.toString('utf-8'));
+  const bundle = JSON.parse(bundleFile.toString('utf-8')) as sigstore.Bundle;
```
feelepxyz, line 95: @bdehamer ops spotted the ts-reset lib only meant for apps not libs, so would reset the consuming app code 🤦 removed it from this PR.

https://github.com/total-typescript/ts-reset#use-ts-reset-in-applications-not-libraries

tsconfig.json
```diff
@@ -8,5 +8,17 @@
...
+  "exclude": [
+    "node_modules",
+    "**/__tests__"
+  ],
+  "strict": true,
+  "exactOptionalPropertyTypes": true,
+  "noFallthroughCasesInSwitch": true,
+  "noImplicitOverride": true,
+  "allowUnreachableCode": false,
+  "forceConsitentCasingInFileNames": true,
```
bdehamer, line 20:
```suggestion
  "forceConsistentCasingInFileNames": true,
```

bdehamer (changes requested):
I agree w/ the the spirit of this change (let's be as strict as possible!), but I think the various config options you added in "tsconfig.json" need to be in the `compilerOptions` section.

src/verify.ts
```diff
@@ -39,7 +39,7 @@ export class Verifier {
     options: sigstore.RequiredArtifactVerificationOptions,
     data?: Buffer
   ): void {
-    this.verifyArtifactSignature(bundle, options, data);
+    this.verifyArtifactSignature(bundle, data);
```
feelepxyz, line 42: Unused options param

tsconfig.json
feelepxyz: ops!

src/tlog/verify/body.ts
```diff
@@ -26,7 +26,9 @@ export function verifyTLogBody(
   bundleContent: sigstore.Bundle['content']
 ): boolean {
   const { kind, version } = entry.kindVersion;
-  const body: EntryKind = JSON.parse(entry.canonicalizedBody.toString('utf8'));
+  const body = JSON.parse(
+    entry.canonicalizedBody.toString('utf8')
+  ) as EntryKind;
```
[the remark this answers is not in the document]
feelepxyz: Yeah seems like both work because `JSON.parse` returns `any`, but with the reset it changed it to `unknown` so had to explicitly cast it. Changed it back to how it was now that we're not including the reset lib 👍

tsconfig.json
```diff
@@ -8,5 +8,17 @@
   "include": [
     "src/**/*"
   ],
-  "exclude": ["node_modules", "**/__tests__"]
+  "exclude": [
+    "node_modules",
+    "**/__tests__"
+  ],
+  "strict": true,
+  "exactOptionalPropertyTypes": true,
```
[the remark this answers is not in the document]
feelepxyz: You're right, removed this for now.

bdehamer (approved):
Nice. Thanks for doing this.

PR #324, opened 2023-03-01
setup changeset workflow for package publishing

bdehamer:
#### Summary
Set-up [changesets](https://github.com/changesets/changesets) to manage package/release publishing.

In place of our existing workflow which relies on the manual creation of a GH release to trigger package publication, we're going to use the `changesets` workflow. Under this model, each of our changesets are queued in a PR and merging that PR triggers all of the following:

* Publication of the package to npmjs
* Creation of a GH release with release notes based on the contents of the changesets
* Amendment of a `CHANGELOG.md` file based on the contents of the changesets


package.json
```diff
@@ -21,6 +22,9 @@
     "type": "git",
     "url": "git+https://github.com/sigstore/sigstore-js.git"
   },
+  "publishConfig": {
+    "provenance": true
+  },
```
bdehamer, line 27: The invocation of `npm publish` happens as part of the `changeset publish` [command](https://github.com/changesets/changesets/blob/main/packages/cli/src/commands/publish/npm-utils.ts#L177-L183) and doesn't give us an opportunity to inject additional flags (we should probably consider a PR to the changeset project with this change).

Since we can't pass the `--provenance` flag to the `publish` command directly we need to set it here (another option is to check-in a `.npmrc` file with this flag set). 

One nice side-effect of this approach is that it prevents anyone from accidentally publishing from their local machine (since you can't publish w/ provenance unless your running on GH Actions).

bdehamer: I see now that this was already being discussed here: https://github.com/gh-community/npm-provenance-private-beta-community/issues/14

feelepxyz: Yes good idea using `publishConfig`! Getting lost in all the different ways you can configure npm.. 😄

feelepxyz (dismissed):
Nice!

feelepxyz:
There's also this bot we can install to comment on PRs that don't have changesets: https://github.com/apps/changeset-bot

We could also fail CI if changesets are missing: https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md#how-do-i-want-to-ensure-pull-requests-have-changesets

feelepxyz:
If we push tags we can probably also use Myles PR and create a release: https://github.com/sigstore/sigstore-js/pull/262

feelepxyz:
> @feelepxyz The changesets CLI doesn't push tags, but the action [does](https://github.com/changesets/action/blob/595655c3eae7136ff5ba18200406898904362926/src/run.ts#L118-L121) -- it automatically creates a release with an associated tag (I'm already using this on the `tuf-js` repo -- here's an example of a [release](https://github.com/theupdateframework/tuf-js/releases/tag/tuf-js%401.1.1)).

oh awesome!

feelepxyz (approved):
🥳

Issue #301, opened 2023-02-15
json file missing


laurentsimon:
I get this error when using the library and `ncc` to compile my project: `::error::Cannot find module '../../store/public-good-instance-root.json'`

Problem with tuf-js?

Any pointers? happy to send a PR, but I'm not sure where this file is looked up.

asraa:
This seems like a similar problem to the rekor.pub being distributed - for Ian and Laurent - this file is likely the initial trusted root after which the Rekor/Fulcio key material is fetched from

asraa:
I think this is where it is being read: https://github.com/sigstore/sigstore-js/blob/6ccf8d04a6b8c12e0a2457f217325a022dff7b76/src/tuf/index.ts#L63

There's a small indirection here that probably made codesearch hard. The `public-good-instance-root.json` is fetched from its definition in this `map.json` file that describes the TUF roots: https://github.com/sigstore/sigstore-js/blob/6ccf8d04a6b8c12e0a2457f217325a022dff7b76/store/map.json

laurentsimon:
Thanks Asra, indeed I just did a lazy `grep` which returned nothing :-)

The fix for the rekor.pub was to use `require.resolve` (see https://github.com/sigstore/sigstore-js/pull/216/files) but the code here already uses that.

I wonder whether the indirection makes `ncc` not realize that it needs to copy the file over, or something to this effect. The hard-coded lookup https://github.com/sigstore/sigstore-js/blob/6ccf8d04a6b8c12e0a2457f217325a022dff7b76/src/tuf/index.ts#L46 seems to be working just fine.

bdehamer:
Haven't used `ncc` before, but it does seem that the dynamic nature of the resolution of the `public-good-instance-root.json` makes it so that `ncc` doesn't realize that this file needs to be packed.

After I ran `ncc` on the `sigstore-js` project, I saw that the `require.resolve` for the "map.json" file got re-written to:
```javascript
const mapSrc = __nccwpck_require__.ab + "map.json";
```
while the statement for the "public-good-instance-root.json" was left as:
```javascript
const tufRootSrc = require.resolve(`../../store/${name}-root.json`);
```
Definitely open to suggestions for how to make this more `ncc`-friendly.

ianlewis:
Based on https://github.com/vercel/ncc/issues/74#issuecomment-442836858 we might have better luck with ncc static analysis if the require took the following form. I guess it then could know that anything under `../../store/` should be included?

[...]

PR #330, opened 2023-03-03
Refactor TUF bootstrap logic

bdehamer:
#### Summary
This is major refactor of the logic we use to bootstrap the local TUF cache. It accomplishes a few different things:

1. The location of the local TUF cache has been updated to use platform-specific app data directories:
    - Windows: `C:\Users\USERNAME\AppData\Local\sigstore-js\Data`
    - Mac: `~/Library/Application Support/sigstore-js`
    - Linux: `~/.local/share/sigstore-js`
    
 1. The public interface now exposes `tufMirrorURL` and `tufRootPath` parameters (currently just on the `verify` function, but these will be added to the `sign*` functions at some point). These params allow the consumer to specify a custom value for the remote TUF repository URL and a path to the "root.json" file which should be used to bootstrap the TUF repository.\
 \
These options are similar to the `--mirror` and `--root` flags exposed by the `cosign initialize` command used to initialize the Sigstore TUF root.\
 \
Most users will simply use the defaults, but exposing these options will make it possible to point the client at different Sigstore instances or write tests against mocked Sigstore services.

1. The `map.json` file ([TAP-4](https://github.com/theupdateframework/taps/blob/master/tap4.md)) had been removed and we're just assuming that the public-good Sigstore instance is the default instance. The indirection introduced by the `map.json` file was overkill for our needs and was causing significant problems (#301) for users trying to pack their project w/ [ncc](https://github.com/vercel/ncc).\
 \
In place of the `map.json` we're going to write a simple `remote.json` file to the TUF cache which contains the URL for the TUF repository (this is the same pattern used by `cosign`).
    ```json
    {"mirror":"https://sigstore-tuf-root.storage.googleapis.com"}
    ```
    We can re-introduce the `map.json` in the future if we have a legitimate use case where we need to resolve targets across multiple TUF repositories.

The new layout of the local TUF cache will look something like this:
```
/Users/jdoe/Library/Application Support/sigstore-js/
├── remote.json
├── root.json
├── snapshot.json
├── targets
│   ├── ctfe.pub
│   ├── ctfe_2022.pub
│   ├── fulcio.crt.pem
│   ├── fulcio_intermediate_v1.crt.pem
│   ├── fulcio_v1.crt.pem
│   └── rekor.pub
├── targets.json
└── timestamp.json
```


src/__tests__/util/appdata.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe('appDataPath', () => {
+  const appName = 'sigstore-js';
+
+  it('it includes the app name', () => {
+    expect(appDataPath(appName)).toContain(appName);
+  });
+
+  // Too complicated to mock out process.platform, os.homedir, and path.join
+  // for each test case. Instead, we'll just test the happy path for each
```
feelepxyz, line 11: 👍

src/tuf/index.ts
```diff
@@ -19,77 +19,81 @@ import { Updater } from 'tuf-js';
...
+  if (!fs.existsSync(targetsPath)) {
+    fs.mkdirSync(targetsPath);
+  }
 
-  if (!fs.existsSync(mapDest)) {
-    const mapSrc = require.resolve('../../store/map.json');
-    fs.copyFileSync(mapSrc, mapDest);
+  if (!fs.existsSync(cachedRootPath)) {
+    const initialRootPath = rootPath || require.resolve(DEFAULT_ROOT_PATH);
+    fs.copyFileSync(initialRootPath, cachedRootPath);
```
feelepxyz, line 61: Will this work with `ncc`?

bdehamer: Yeah, since the `require.resolve` points to a static path now, ncc detects the file access and re-writes this line:

```javascript
if (!fs_1.default.existsSync(cachedRootPath)) {
    const initialRootPath = rootPath || __nccwpck_require__.ab + "public-good-instance-root.json";
    fs_1.default.copyFileSync(initialRootPath, cachedRootPath);
}
```

The use of `__nccwpck_require__.ab` means that ncc found that file and copied it into the distribution bundle.

src/tuf/index.ts
```diff
@@ -19,77 +19,81 @@ import { Updater } from 'tuf-js';
...
 
 export async function getTrustedRoot(
-  cacheDir: string
+  cachePath: string,
+  options: TUFOptions = {}
 ): Promise<sigstore.TrustedRoot> {
-  initTufCache(cacheDir);
-  const repoMap = initRepoMap(cacheDir);
+  initTufCache(cachePath, options.rootPath);
+  const mirror = initRemoteConfig(cachePath, options.mirrorURL);
```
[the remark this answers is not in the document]
bdehamer: Good call-out. I did some refactoring that makes things more readable.

feelepxyz (dismissed):

PR #358, opened 2023-03-17
make oauth redirect URL configurable

### LINKED ISSUE #306: As a npx sigstore user, I want to specify OIDC callback port in environment variable.
**Description**

similar issue : random callback uri port  · sigstore/cosign https://github.com/sigstore/cosign/issues/1105

workaround

```sh
#!/bin/sh
FOO_CONTAINER=foo1234

docker build -t $FOO_CONTAINER - <<'EOF'
FROM node:18.14-alpine3.17
RUN apk add --no-cache jq openssl curl coreutils
WORKDIR /app
RUN npm install sigstore@v1.0.0
EOF

docker run -i --rm -p 18089:18089 --entrypoint=sh $FOO_CONTAINER <<'EOF'
npm version
# Added environment variable for OIDC callback port and turn xdg-open to echo
cd /app/node_modules/sigstore/dist/identity/
sed -i.bak 's/listen(0/listen(Number(process.env.OIDC_CALLBACK_PORT) || 0/ ; s/xdg-open/echo/' oauth.js
diff oauth.js.bak oauth.js
cd /app
echo "content to be signed" > artifact
export OIDC_CALLBACK_PORT=18089
npx sigstore sign artifact > bundle.sigstore
cat bundle.sigstore | jq
cat bundle.sigstore | jq --raw-output '.verificationMaterial.x509CertificateChain.certificates[0].rawBytes' | \
  base64 -d | openssl x509 -inform der -text
LOGINDEX=`cat bundle.sigstore | jq --raw-output .verificationMaterial.tlogEntries[0].logIndex`
curl -s https://rekor.sigstore.dev//api/v1/log/entries?logIndex=$LOGINDEX | jq --raw-output '.[].body' | \
  base64 --decode | jq
npx sigstore verify bundle.sigstore artifact
EOF
```

diff output

```diff
--- oauth.js.bak
+++ oauth.js
@@ -45,7 +45,7 @@
         const sockets = new Set();
         // Start server and wait till it is listening
         await new Promise((resolve) => {
-            server.listen(0, resolve);
+            server.listen(Number(process.env.OIDC_REDIRECT_PORT) || 0, resolve);
         });
         // Keep track of connections to the server so we can force a shutdown
         server.on('connection', (socket) => {
@@ -160,7 +160,7 @@
                     open = 'open';
                     break;
                 case 'linux' || 'freebsd' || 'netbsd' || 'openbsd':
-                    open = 'xdg-open';
+                    open = 'echo';
                     break;
                 case 'win32':
                     open = 'start';
```

bdehamer:
Closes #306 

#### Summary
Updates the OAuth identity provider to allow for a configurable `redirectURL`. Some authorization servers require a pre-configured redirect URL so using a dynamic port doesn't work. The `redirectURL` option allows a user to specify a specific host/port for listening for the OAuth redirect. If no value is specified it falls back to the previous behavior of picking a random/available port.

#### Release Note
* Adds new `redirectURL` config option when signing with an OAuth identity provider

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -63,6 +63,7 @@ function printUsage() {
 const signOptions = {
   oidcClientID: 'sigstore',
   oidcIssuer: 'https://oauth2.sigstore.dev/auth',
+  oidcRedirectURL: process.env.OIDC_REDIRECT_URL,
   rekorURL: sigstore.DEFAULT_REKOR_URL,
 };
 
--- a/.changeset/nasty-bees-switch.md
+++ b/.changeset/nasty-bees-switch.md
@@ -1,0 +1,5 @@
+---
+'sigstore': minor
+---
+
+Adds new `redirectURL` config option when signing with an OAuth identity provider
--- a/src/__tests__/identity/oauth.test.ts
+++ b/src/__tests__/identity/oauth.test.ts
@@ -28,7 +28,7 @@ jest.mock('child_process', () => ({
 describe('OAuthProvider', () => {
   const baseURL = 'http://localhost:8080';
   const issuer = new Issuer(baseURL);
-  const subject = new OAuthProvider(issuer, 'sigstore', '');
+  const subject = new OAuthProvider({ issuer, clientID: 'sigstore' });
 
   const mockedExec = jest.mocked(child_process.exec);
 
--- a/src/identity/index.ts
+++ b/src/identity/index.ts
@@ -27,12 +27,18 @@ import { Provider } from './provider';
  * @param clientSecret Client secret for the issuer (optional)
  * @returns {Provider}
  */
-function oauthProvider(
-  issuer: string,
-  clientID: string,
-  clientSecret?: string
-): Provider {
-  return new OAuthProvider(new Issuer(issuer), clientID, clientSecret);
+function oauthProvider(options: {
+  issuer: string;
+  clientID: string;
+  clientSecret?: string;
+  redirectURL?: string;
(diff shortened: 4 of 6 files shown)


src/cli/index.ts
```diff
@@ -63,6 +63,7 @@ function printUsage() {
 const signOptions = {
   oidcClientID: 'sigstore',
   oidcIssuer: 'https://oauth2.sigstore.dev/auth',
+  oidcRedirectURL: process.env.OIDC_REDIRECT_URL,
```
feelepxyz, line 66: Do we want to set this using env or should it be a [flag like cosign](https://github.com/sigstore/cosign/blob/ce8177ce72c54bfa55b8de2e65a97d3e608ba4f1/doc/cosign_sign.md?plain=1#L88)? 🤔 Also found [this env var in gitsign](https://github.com/sigstore/gitsign/blob/2509c5c5f931b5de2cc9ca2449d7e309a65a2ec9/internal/config/config_test.go#L71).

feelepxyz, line 66: Should we prefix the env var with `SIGSTORE`?

```suggestion
  oidcRedirectURL: process.env.SIGSTORE_OIDC_REDIRECT_URL,
```

feelepxyz: Yeah good point, I like this idea of publishing the CLI as a separate package. Would also mean we start from scratch on the interface.

feelepxyz (approved):
Minor comment otherwise 👍

PR #371, opened 2023-03-23
Automate creating changesets for dependabot PRs

feelepxyz:
Hacked together a action to autogenerate changesets for dependabot PRs. I started out thinking it would be super easy to just invoke the changesets cli from a custom workflow but it's not possible to pass in the necessary args, and realised it's doing a lot more than just taking a summary and semver update type so ended up with writing this https://github.com/feelepxyz/dependabot-changesets 

Not sure it's a great idea yet 😆  Thoughts @bdehamer?

### DIFF
--- a/.github/workflows/dependabot-changesets.yml
+++ b/.github/workflows/dependabot-changesets.yml
@@ -1,0 +1,15 @@
+name: "Dependabot Changesets"
+
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  changesets:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'sigstore' &&  github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot Changesets
+        uses: feelepxyz/dependabot-changesets@b91eb2125cabe58b66aa7f18d3345855108d6e21 # v1


.github/workflows/dependabot-changesets.yml
```diff
@@ -0,0 +1,15 @@
...
+  contents: write
+  pull-requests: write
+
+jobs:
+  changesets:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'sigstore' &&  github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot Changesets
+        uses: feelepxyz/dependabot-changesets@b91eb2125cabe58b66aa7f18d3345855108d6e21 # v1
```
feelepxyz, line 15: Here's an example: https://github.com/feelepxyz/test-dependabot/pull/3/files 

I opted to always create `patch` updates instead of inferring from the semver update of the depenedency or something else clever. Thinking it should be easy enough to just edit the created changelog in the pr files tab if necessary.

feelepxyz: @bdehamer updated it to create empty changesets for indirect/dev dependencies: https://github.com/feelepxyz/dependabot-changesets/blob/main/src/main.ts

Does this make sense to you? 

I also updated it to now create a separate changelog entry for each changed package if there are multiple and create a summary for the packages updated dependencies: https://github.com/feelepxyz/dependabot-changesets/blob/main/src/main.ts#LL146C7-L146C22 (I suspect there might be cases where this does the wrong thing as I haven't tested this at all on monorepos)

feelepxyz, line 15:
```suggestion
        uses: feelepxyz/dependabot-changesets@0d623339356e43f9b962da61329803e33fffa9d1 # v1
```

.github/workflows/dependabot-changesets.yml
```diff
@@ -0,0 +1,15 @@
...
+  contents: write
+  pull-requests: write
+
+jobs:
+  changesets:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'sigstore' &&  github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot Changesets
+        uses: feelepxyz/dependabot-changesets@0d623339356e43f9b962da61329803e33fffa9d1 # v1
```
feelepxyz, line 15:
```suggestion
        uses: feelepxyz/dependabot-changesets@2d6ec3d11f39d39682c6ea53be1bd5faeb093897 # v1
```

bdehamer (approved):
Awesome!

.github/workflows/dependabot-changesets.yml
```diff
@@ -0,0 +1,15 @@
...
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  changesets:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'sigstore' &&  github.actor == 'dependabot[bot]' }}
```
bdehamer, line 12: What is the `github.repository_owner == 'sigstore'` protecting here?

feelepxyz: > What is the `github.repository_owner == 'sigstore'` protecting here?

So it doesn't run on forks but dunno if really worth it. 

Yeah maybe could also provide a reusable workflow that invokes itself?

feelepxyz: or invoke the reusable workflow from the action? 😄

bdehamer (approved):

bdehamer (dismissed):

bdehamer (approved):

PR #388, opened 2023-04-04
verify cert extensions w/ UTF8String encoding

bdehamer:
#### Summary
Updates the certificate extension verification logic to support the newer-style Fulcio extension encoding (where the extension value is encoded as a `UTF8String` before it's added to the extension structure).


feelepxyz (approved):
👍

haydentherapper:
Nice!

PR #414, opened 2023-04-18
Update CLI sign cmd to output a link to search

feelepxyz:
Update the CLI sign command to output a link to search.sigstore.dev when the rekor url is set to the public good instance. 

I mainly wanted this for demo purposes as it makes it easy to just cmd click on the link and immediately get something useful instead of the rekor api entry.


bdehamer (dismissed):
Nice!

bdehamer (approved):

PR #425, opened 2023-04-20
release fixup

bdehamer:
#### Summary
Merging the last [release PR](https://github.com/sigstore/sigstore-js/pull/424) didn't trigger a publish. Hoping that merging this will cause `changeset` to detect that the current version hasn't been published.


feelepxyz:
Trying to figure out what went wrong here. Checking this successful publish: https://github.com/sigstore/sigstore-js/actions/runs/4754341887/jobs/8447049392 
This one says `No changesets found, attempting to publish any unpublished packages to npm`

The one that failed: https://github.com/sigstore/sigstore-js/actions/runs/4759623510/jobs/8459576890 
Says `All changesets are empty; not creating PR` 

I think the issue is actually that there is still an empty changeset on main: https://github.com/sigstore/sigstore-js/blob/main/.changeset/wild-rings-brake.md

This is strange, because [the empty changeset was merged in 17 hours ago](https://github.com/sigstore/sigstore-js/pull/423) and the [release that failed was merged in 11 hours ago](https://github.com/sigstore/sigstore-js/pull/424), I would have expected it to be deleted in the version pr.

Digging further, it seems like these dependabot PRs, [when merged into main](https://github.com/sigstore/sigstore-js/commit/b6d920a2a3fe2638a09700467a817c00cf1eccde) don't trigger builds.

Pretty sure this is because github actions comitted the change using the `GITHUB_TOKEN` which [won't trigger additional workflow runs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow).

--

TL'DR: When GH actions automerges dependabot prs, these don't get included in `Version Packages` Pull Requests that are already open, so don't get deleted, and when left on `main` prevent a new release from going out.

Possible solutions:
- Use a PAT instead of `GITHUB_TOKEN`  to do the auto merging
- Stop doing auto-merging with GH?
- Maybe others I haven't thought of

feelepxyz:
@bdehamer I think the short term fix here is to delete the changeset on main

feelepxyz (approved):

PR #427, opened 2023-04-22
allow for empty tlogEntries in bundle

bdehamer:
#### Summary
Updates the `toDSSEBundle` and `toMessageSignatureBundle` functions to allow for an optional tlog entry. This supports use cases where a user may opt-out of uploading data to the transparency log.


src/types/sigstore/index.ts
```diff
@@ -180,15 +188,18 @@ function toTransparencyLogEntry(entry: Entry): TransparencyLogEntry {
...
+}: {
+  signature: SignatureMaterial;
+  tlogEntry?: Entry;
+}): VerificationMaterial {
   return {
     content: signature.certificates
       ? toVerificationMaterialx509CertificateChain(signature.certificates)
       : toVerificationMaterialPublicKey(signature.key.id || ''),
-    tlogEntries: [toTransparencyLogEntry(entry)],
+    tlogEntries: tlogEntry ? [toTransparencyLogEntry(tlogEntry)] : [],
```
bdehamer, line 202: `tlogEntries` may now be an empty array

src/types/sigstore/index.ts
```diff
@@ -120,25 +121,32 @@ export function isVerifiableTransparencyLogEntry(
...
-  ): Bundle => ({
+export function toDSSEBundle({
+  envelope,
+  signature,
+  tlogEntry,
+}: {
+  envelope: Envelope;
+  signature: SignatureMaterial;
+  tlogEntry?: Entry;
+}): Bundle {
```
bdehamer, line 132: Switch from ordered arguments an object so we can make the `tlogEntry` optional (and add support for an optional `timestampEntry` in the future)

PR #428, opened 2023-04-23
separate bundle assembly from tlog submission

bdehamer:
#### Summary
De-couples the assembly of the Sigstore bundle from the TLog submission logic. We need to allow for scenarios where an artifact is signed but NOT submitted the transparency log -- this change lays the groundwork for the eventual introduction of a `tlogUpload` flag.


feelepxyz (dismissed):
Nice!

feelepxyz (approved):

PR #437, opened 2023-04-26
add tlogUpload option

bdehamer:
#### Summary
Introduces a new `tlogUpload` option to the `sign` and `attest` functions which can be used to control whether the generated signature is uploaded to the transparency log. Defaults to `true`.

### DIFF
--- a/src/__tests__/verify.test.ts
+++ b/src/__tests__/verify.test.ts
@@ -161,6 +161,61 @@ describe('Verifier', () => {
             );
           });
         });
+
+        describe('when there are no tlog entries in the bundle', () => {
+          const bundle = sigstore.bundleFromJSON(
+            bundles.dsse.valid.withNoTLogEntries
+          );
+
+          describe('when tlog verification is disabled', () => {
+            const opts: sigstore.RequiredArtifactVerificationOptions = {
+              ...options,
+              tlogOptions: {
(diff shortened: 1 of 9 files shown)


feelepxyz (commented):
Changes look good, was only wondering if we need/want some tests to make sure we can verify these bundles or does this intentionally not work? Should the verification also take an option to verify the tlog entry to make sure it exists so you couldn't just bypass this by messing with the bundle?

bdehamer:
@feelepxyz I added an explicit test for verifying a bundle with no tlog entries. There are actually two different ways to get this to pass verification: 

1) Set the `tlogOptions.disabled` verification option to `true`
2) Set the `tlogOptions.threshold` verification option to `0`

src/__tests__/verify.test.ts
```diff
@@ -161,6 +161,61 @@ describe('Verifier', () => {
...
+              const opts: sigstore.RequiredArtifactVerificationOptions = {
+                ...options,
+                tlogOptions: {
+                  disable: false,
+                  threshold: 1,
+                  performOnlineVerification: false,
+                },
+              };
+
+              it('throws an error', () => {
```
feelepxyz, line 211: Nice!

feelepxyz (approved):
👍

PR #448, opened 2023-05-03
add client for making tsa requests

bdehamer:
#### Summary
First step toward support for requesting timestamps from the [Timestamp Authority](https://github.com/sigstore/timestamp-authority).

This just adds a client for interacting w/ the TSA API. Plenty more work to do here to fully enable this feature.


src/external/tsa.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ua } from '../util';
+import { checkStatus } from './error';
+
+export interface TimestampRequest {
+  artifact: string;
+  hashAlgorithm: string;
+  certificates?: boolean;
+  nonce?: number;
+  tsaPolicyOID?: string;
+}
```
bdehamer, line 26: @malancas would love to get your eyes on this just to validate the "optionality" of the fields in the request.

bdehamer, line 26: @malancas would love to get your eyes on this just to validate the "optionality" of the fields in the request.

malancas: The optionality of these fields looks correct. Only the artifact and hash algorithm are required.

bdehamer:
Closing until we work through some issues w/ the TSA API interface

PR #452, opened 2023-05-04
omit Fulcio root/intermediate certs from bundle

bdehamer:
#### Summary
Per https://github.com/sigstore/protobuf-specs/issues/75, updates the emitted bundle to exclude the Fulcio root and intermediate certificates.


feelepxyz (approved):
LGTM 👍

PR #453, opened 2023-05-04
only consider leaf cert in bundle for verification

bdehamer:
#### Summary
Per https://github.com/sigstore/protobuf-specs/issues/75, updates the certificate verification logic to ignore any root/intermediate certificates which may be present in the bundle. The certificates needed to complete the chain MUST come from the trusted root.

The certificate chain verification logic was refactored so that it's not even possible to pass in more than a single, untrusted certificate.

This change is fully backward compatible with all existing, valid bundles.


.changeset/cold-cobras-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'sigstore': patch
+---
+
+Updates certificate verification to ignore all but the first certificate present in the bundle
```
feelepxyz, line 5: nit
```suggestion
Updates certificate verification to ignore all but the first leaf certificate present in the bundle
```

feelepxyz (approved):
Nice one!

PR #459, opened 2023-05-05
pathlen constraint

bdehamer:
#### Summary
Updates the certificate chain verification logic to obey the pathlen constraint (if present) of any CA certificates. This helps protect against attacks where someone is able to provision an unauthorized intermediate certificate and then attempts to use it to issue their own leaf certs.

The Fulcio intermediate certificate has a `pathlen:0` constraint (meaning that no additional intermediates may be present in the chain). This update ensures that we enforce that constraint.

There is one other, related change present here related to the certificate chain which is returned after it has been verified. Previously, this chain had two copies of the root cert (since the root cert is self-signed, its issuer is itself) but this may be confusing (and makes the pathlen verification algorithm difficult) so I've removed the duplicate root from the returned chain.


src/x509/verify.ts
```diff
@@ -192,6 +194,25 @@ class CertificateChainVerifier {
...
+      // If the certificate is a CA, check the path length
+      if (cert.extBasicConstraints?.isCA) {
+        const pathLength = cert.extBasicConstraints.pathLenConstraint;
+
+        // The path length, if set, indicates how many intermediate
+        // certificates (NOT including the leaf) are allowed to follow. The
+        // pathLength constraint of any intermediate CA certificate MUST be
+        // greater than or equal to it's own depth in the chain (with an
+        // adjustment for the leaf certificate)
+        if (pathLength !== undefined && pathLength < i - 1) {
```
feelepxyz, line 211: Is the `pathLenConstraint` attribute always set on CA cert objects and when not explicitly it's`undefined`?

Is the path length set on the intermediate and the root? What are the path length constraints set to on the fulcio certs? 

Just walking through the logic in my head, seems like `i - 1` will be `0` for the intermediate and `1` for the root. So the path length constraint needs to be `<= -1` to fail for the intermediate (`>= 0` to succeed), and `<= 0` to fail for the root (`>= 1` to succeed).

bdehamer: https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.9

feelepxyz (approved):
🚢

PR #468, opened 2023-05-09
support for requesting tsa timestamps

bdehamer:
#### Summary
Adds support for requesting a signed timestamp over the artifact signature which is added to the bundle. Both the `sign` and `attest` functions now support a `tsaURL` option which specifies the URL for the [Timestamp Authority API](https://github.com/sigstore/timestamp-authority) which should be used to request the timestamp. If no value is specified, a timestamp will not be requested.


README.md
```diff
@@ -40,6 +40,7 @@ necessary to verify the signature.
 * `options` `<Object>`
   * `fulcioURL` `<string>`: The base URL of the Fulcio instance to use for retrieving the signing certificate. Defaults to `'https://fulcio.sigstore.dev'`.
   * `rekorURL` `<string>`: The base URL of the Rekor instance to use when adding the signature to the transparency log. Defaults to `'https://rekor.sigstore.dev'`.
+  * `tsaURL` `<string>`: The base URL of the Timestamp Authority instance to use when requesting a signed timestamp. If omitted, no timestamp will be requested.
```
feelepxyz, line 43: Should we name this `tsaServerURL` to match cosign? https://github.com/sigstore/cosign/blob/cc969943428c109904d6a1888eb44c1576769cbd/cmd/cosign/cli/attest/attest.go#L83
```suggestion
  * `tsaServerURL` `<string>`: The base URL of the Timestamp Authority instance to use when requesting a signed timestamp. If omitted, no timestamp will be requested.
```

src/external/tsa.ts
```diff
@@ -0,0 +1,59 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        'User-Agent': ua.getUserAgent(),
+      },
+    });
+    this.baseUrl = options.baseURL;
+  }
+
+  public async createTimestamp(request: TimestampRequest): Promise<Buffer> {
+    const url = `${this.baseUrl}/api/v1/timestamp`;
```
feelepxyz, line 49: Is this in a spec somewhere or how do we know it's always `/api/v1/timestamp`?

feelepxyz: Seems like a sigstore thing from googling around: https://github.com/sigstore/timestamp-authority

Seems fine, but guess this means any sigstore client only supports the "sigstore" flavoured TSA setup.

feelepxyz (commented):
Looks great to me! Just a minor nit on renaming to `tsaServerURL`

README.md
bdehamer: Updated

src/external/tsa.ts
bdehamer: Yeah, this endpoint and the request format we're using is Sigstore-specific

feelepxyz (approved):

PR #469, opened 2023-05-09
expose fetch options

bdehamer:
#### Summary
Removes the hardcoded `retry` and `timeout` fetch values and exposes top-level configuration options which can be used by the consumer. These values are propagated down to the individual clients used to interact w/ Fulcio, Rekor and TUF (will update with TSA support once that PR has merged).

This allows us to easily inherit the same values configured for the npm CLI. Also gives us the ability to set specific values during testing to ensure we don't waste time retrying requests when running unit tests.


.changeset/twelve-moles-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'sigstore': minor
+---
+
+Expose new `retry` and `timeout` options to control fetch behavior when errors occure
```
feelepxyz, line 5: Just looking at the documentation for [make-fetch-happen retry](https://github.com/npm/make-fetch-happen#opts-retry) and kinda feeling like we're tying our interface to the underlying http lib which might change in future. 

Thoughts on just exposing `retries: number` and `timeout: number`? We could then have freedom to tune the defaults that make sense for the sigstore public good, as we'll have good insight into which errors will be retriable, what the timeouts/backoffs should be etc.

src/tuf/index.ts
```diff
@@ -115,13 +117,33 @@ function initRemoteConfig(rootDir: string, mirrorURL: string): RemoteConfig {
...
+  // convert the boolean and object options to a number.
+  if (typeof options.retry !== 'undefined') {
+    if (typeof options.retry === 'number') {
+      config.fetchRetries = options.retry;
+    } else if (typeof options.retry === 'object') {
+      config.fetchRetries = options.retry.retries;
+    } else if (options.retry === true) {
+      config.fetchRetries = 1;
+    }
+  }
```
feelepxyz, line 140: Looks like this could also be simplified if we just exposed `retries: number`

src/types/fetch.ts
```diff
@@ -0,0 +1,8 @@
+import type { MakeFetchHappenOptions } from 'make-fetch-happen';
+
+export type Retry = MakeFetchHappenOptions['retry'];
+
+export type FetchOptions = {
+  retry?: Retry;
```
feelepxyz, line 6:
```suggestion
  retries?: number | undefined;
```

.changeset/twelve-moles-appear.md
bdehamer: Yeah, I agree it is a bit "leaky" (that isn't even a `make-fetch-happen` thing, it's really the interface for the underlying `retry` library). The reason I opted to go with this approach is that this is also the same interface that the npm CLI uses internally (and exposes through some [command line args](https://github.com/npm/cli/blob/f06469607b80faf72eb897ec3e33deebc6dc10fc/lib/utils/config/definitions.js#L654-L709)).  The public interface for `npm-registry-fetch` also exposes the same [retry option](https://github.com/npm/npm-registry-fetch#-optsretry).

Since our primary use case has us sandwiched between npm CLI and `make-fetch-happen` it seemed easiest to just adopt the data structure which those two had already agreed upon -- that way, any future changes will just flow through `sigstore-js` without us having to do anything.

feelepxyz: > Since our primary use case has us sandwiched between npm CLI and make-fetch-happen it seemed easiest to just adopt the data structure which those two had already agreed upon -- that way, any future changes will just flow through sigstore-js without us having to do anything.

Yeah these are good point, being tightly integrated in npm cli and will probably continue to be the main use-case for the foreseeable. Part of me still leans toward the simple option and ignoring all the other retry options but 🤷

feelepxyz (approved):
SGTM 👍

feelepxyz (approved):

PR #473, opened 2023-05-10
enable dependabot grouped updates

bdehamer:
#### Summary
Adding the most-frequently updated dev dependencies to a Dependabot update group.


.github/dependabot.yml
```diff
@@ -6,6 +6,13 @@ updates:
     interval: daily
   versioning-strategy: auto
   open-pull-requests-limit: 10
+  groups:
+    dev-dependencies:
+      patterns:
+        - "@types/*"
```
feelepxyz, line 12: Ah there's no way of saying things like `dependency-type: development`?

PR #474, opened 2023-05-11
convert to monorepo

bdehamer:
#### Summary
Reorganizes the layout of the repository to support multiple npm packages. For now, `sigstore` is the only package in the repository, but there are plans to add a CLI and some other distinct packages that we want to be able to publish separately from the core Sigstore client.

Includes a reorganization of the TypeScript and Jest configuration files so that they can easily used across multiple packages without duplicating the configuration.


.github/workflows/ci.yml
```diff
@@ -59,6 +59,8 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
         cache: npm
+    - name: Update npm
+      run: npm install -g npm@8
```
bdehamer, line 63: For our Node 14 tests we need to make sure we're using `npm` v8 to support the "workspaces" feature in package.json

feelepxyz: Does npm 9 still run tests on 14?

README.md
```diff
@@ -1,159 +1,10 @@
...
 
-A JavaScript library for generating and verifying Sigstore signatures. One of
-the intended uses is to sign and verify npm packages but it can be used to sign
-and verify any file.
+JavaScript libraries for interacting with [Sigstore][6] services.
 
-## Features
+## Packages
 
-* Support for signing using an OpenID Connect identity
```
feelepxyz, line 9: hm do we want to make these changes to the overall readme? Feels like it's more valuable to focus on the main lib here and maybe link out to the other sub-packages?

feelepxyz: although can also see how we want a readme in the client folder: packages/client/README.md

.github/workflows/ci.yml
feelepxyz: ah yeah looks like it https://github.com/npm/cli/blob/latest/.github/workflows/ci.yml#L70-L71

feelepxyz (approved):
LGTM

README.md
bdehamer: I didn't want to have to maintain duplicate content between two readmes . . . and the one in the client folder is the one that will appear on the [npm page](https://www.npmjs.com/package/sigstore) so it seemed appropriate to have the library usage information there.

PR #476, opened 2023-05-12
split rekor types into dedicated package

bdehamer:
#### Summary
Adds a new package containing the code-generated Rekor types. This takes our existing Rekor codegen script and adds support for generating the top-level request types from the Rekor openapi schema. Moving all these types into a dedicated package that will only be used at dev-time means that the overall `sigstore` package size will shrink.

Will come back after this is merged/published and add the new package as a development dependency to the `sigstore` package.


packages/rekor-types/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  | {
+      apiVersion: '0.0.1';
+      kind: typeof INTOTO_KIND;
+      spec: IntotoV001Schema;
+    }
+  | {
+      apiVersion: '0.0.2';
+      kind: typeof INTOTO_KIND;
+      spec: IntotoV002Schema;
+    };
```
bdehamer, line 26: Note that we still have some hand-rolled types given that the Rekor schema only partially describes the `ProposedEntry` types.

feelepxyz (approved):
Nice!

feelepxyz (approved):

PR #477, opened 2023-05-12
CLI

bdehamer:
#### Summary

First pass at a basic [oclif](https://oclif.io/)-generated CLI. Initially publishing as a 0.0.1 release so we can iterate on the interface.

image

image

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -1,0 +1,107 @@
+# @sigstore/cli
+
+CLI for creating and verifying Sigstore bundles.
+
+# Usage
+<!-- usage -->
+```sh-session
+$ npm install -g @sigstore/cli
+$ sigstore COMMAND
+running command...
+$ sigstore (--version)
+@sigstore/cli/0.0.1 darwin-arm64 node-v18.12.1
+$ sigstore --help [COMMAND]
+USAGE
+  $ sigstore COMMAND
+...
+```
+<!-- usagestop -->
+# Commands
+<!-- commands -->
+* [`sigstore attest FILE`](#sigstore-attest-file)
+* [`sigstore help [COMMANDS]`](#sigstore-help-commands)
+* [`sigstore verify BUNDLE`](#sigstore-verify-bundle)
+
+## `sigstore attest FILE`
+
+attest the supplied file
+
+```
+USAGE
+  $ sigstore attest FILE [--json] [--fulcio-url <value>] [--rekor-url <value>] [--tsa-server-url <value>]
+    [--tlog-upload] [--oidc-client-id <value>] [--oidc-issuer <value>] [--oidc-redirect-url <value>] [-t <value>] [-o
+    <value>]
+
+ARGUMENTS
+  FILE  file to attest
+
+FLAGS
+  -o, --output-file=<value>    write output to file
+  -t, --type=<value>           [default: application/vnd.in-toto+json] type to apply to the DSSE envelope
+  --fulcio-url=<value>         [default: https://fulcio.sigstore.dev] URL to the Sigstore PKI server
+  --oidc-client-id=<value>     [default: sigstore] OIDC client ID for application
+  --oidc-issuer=<value>        [default: https://oauth2.sigstore.dev/auth] OIDC provider to be used to issue ID token
+  --oidc-redirect-url=<value>  OIDC redirect URL
+  --rekor-url=<value>          [default: https://rekor.sigstore.dev] URL to the Rekor transparency log
+  --tlog-upload                whether or not to upload entry to the transparency log
+  --tsa-server-url=<value>     URL to the Timestamping Authority
+
+GLOBAL FLAGS
+  --json  Format output as json.
+
+DESCRIPTION
+  attest the supplied file
+
+EXAMPLES
+  $ sigstore attest
+```
+
+_See code: [dist/commands/attest.ts](https://github.com/sigstore/sigstore-js/blob/v0.0.1/dist/commands/attest.ts)_
+
+## `sigstore help [COMMANDS]`
+
+Display help for sigstore.
+
+```
+USAGE
+  $ sigstore help [COMMANDS] [-n]
+
+ARGUMENTS
+  COMMANDS  Command to show help for.
+
+FLAGS
+  -n, --nested-commands  Include all nested commands in the output.
+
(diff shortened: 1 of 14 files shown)


packages/cli/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "homepage": "https://github.com/sigstore/sigstore-js/tree/main/packages/cli#readme",
+  "bugs": "https://github.com/sigstore/sigstore-js/issues",
+  "oclif": {
+    "bin": "sigstore",
+    "dirname": "sigstore",
+    "commands": "./dist/commands",
+    "plugins": [
+      "@oclif/plugin-help"
+    ],
+    "topicSeparator": " "
```
feelepxyz, line 54: 👍  seems like a more ergonomic sub-command separator over `:` and matches npm cli (had to [look this up](https://oclif.io/docs/topic_separator))

packages/cli/package.json
```diff
@@ -0,0 +1,59 @@
...
+  ],
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "bin": {
+    "sigstore": "./bin/run"
+  },
+  "files": [
+    "/bin",
+    "/dist",
+    "/oclif.manifest.json"
```
feelepxyz, line 19: Does this manifest need to be checked in?

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,51 @@
...
+    const options: Parameters<typeof sigstore.verify>[2] = {
+      tlogThreshold: flags['tlog-threshold'],
+      ctLogThreshold: flags['ctlog-threshold'],
+    };
+
+    const bundle = await fs
+      .readFile(args.bundle)
+      .then((data) => JSON.parse(data.toString()));
+
+    return sigstore.verify(bundle, undefined, options).then(() => {
```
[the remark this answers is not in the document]
feelepxyz: Do we want to allow passing in `data` for `messageSignature` bundles`? Optional third arg?

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default class Verify extends Command {
+  static override description = 'describe the command here';
+  static override examples = [
+    '<%= config.bin %> <%= command.id %> ./bundle.sigstore',
+  ];
+  static override enableJsonFlag = true;
+
+  static override flags = {
+    'tlog-threshold': Flags.integer({
```
feelepxyz, line 13: Presuming you punted on the other options until we decide exactly how to expose these like?
- certificateIssuer?: string;
- certificateIdentityEmail?: string;
- certificateIdentityURI?: string;
- certificateOIDs?: Record<string, string>;
- tufMirrorURL?: string;
- tufRootPath?: string;
- tufCachePath?: string;
- retry
- timeout

packages/cli/src/commands/attest.ts
```diff
@@ -0,0 +1,114 @@
...
+      this.printRekorEntry(flags['rekor-url'], bundle);
+    }
+
+    if (flags['output-file']) {
+      await fs.writeFile(flags['output-file'], JSON.stringify(bundle));
+    } else {
+      this.log(JSON.stringify(bundle));
+    }
+
+    return bundle;
```
feelepxyz, line 96: Does the return do anything when you log?

packages/cli/bin/dev.cmd
```diff
@@ -0,0 +1,3 @@
+@echo off
```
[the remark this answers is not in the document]
bdehamer: The `dev` script allows you to run your CLI without first compiling the TypeScript (just a convenience during development). The `dev.cmd` is some sort of Windows-specific wrapper on top of `dev`.

Both of these are code-generated as part of the oclif set-up.

packages/cli/README.md
```diff
@@ -0,0 +1,107 @@
...
+$ sigstore (--version)
+@sigstore/cli/0.0.1 darwin-arm64 node-v18.12.1
+$ sigstore --help [COMMAND]
+USAGE
+  $ sigstore COMMAND
+...
+```
+<!-- usagestop -->
+# Commands
+<!-- commands -->
```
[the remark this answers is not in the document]
bdehamer: Exactly

packages/cli/package.json
bdehamer: No, it's generated as part of the release process, packaged in the tarball and then immediately deleted.

Really we could omit the manifest altogether -- I believe it's only useful if we wanted this CLI to function as a plugin to some other CLI.

packages/cli/src/commands/verify.ts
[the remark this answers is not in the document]
bdehamer: > Presuming this will exit 1 if verification fails when throwing and then exit 0 on success

Yep.

> Should we add some basic happy/sad path tests around this?

Yeah, I'm gonna come back through and figure out the whole testing strategy for this. There is so much mocking required to do even basic tests here that I decided to punt on this for the first pass.

I'm gonna punt on the messageSignature use case for the moment. I'm wondering if it should be a separate command altogether.

packages/cli/src/commands/attest.ts
bdehamer: If you specify the `--json` flag when you invoke the CLI oclif will suppress anything you write with `log` and instead pretty-print whatever JSON is returned from the command.

packages/cli/src/commands/verify.ts
bdehamer: Yeah, will come back and address all of those later. I feel like the `certificate*` options need some refactoring still and the rest of those probably need to be global options (shared across all the CLI commands) but I think that requires that I extract a shared bass class.

packages/cli/bin/dev.cmd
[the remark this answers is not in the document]
feelepxyz: Ahh gotcha!

packages/cli/src/commands/attest.ts
feelepxyz: Oh nice!

feelepxyz (approved):
🎉

feelepxyz (approved):

feelepxyz (approved):

PR #483, opened 2023-05-15
Fix to tlog-upload flag in CLI

bdehamer:
#### Summary
Fix to the `attest` command in the CLI to allow the `tlog-upload` flag to be negated. Previously, you could only be enabled . . . which is already the default. Now you can set `--no-tlog-upload` to disable the default behavior.


.changeset/tricky-bats-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sigstore/cli': patch
+---
+
+Fix to allow `--tlog-upload` flag to be negated
```
feelepxyz, line 5:
```suggestion
Fix to allow `--tlog-upload` flag to be negated with `--no-tlog-upload`
```

feelepxyz (approved):

PR #484, opened 2023-05-15
handle detached SCT response from Fulcio

bdehamer:
#### Summary
Updates the Fulcio client to properly handle either of the two response types that may come back when requesting a signing certificate.

Depending on how a particular Fulcio instance is configured it may return either:

```javascript
{ signedCertificateEmbeddedSct: { chain: { certificates: [] } }
```
or
```javascript
{ signedCertificateDetachedSct: { chain: { certificates: [] }, signedCertificateTimestamp: '' }
```


feelepxyz (approved):
LGTM!

PR #490, opened 2023-05-16
consume @sigstore/rekor-types in client

bdehamer:
#### Summary
Updates the client to pull all of the Rekor types from the `@sigstore/rekor-types` package.


packages/client/src/types/sigstore/index.ts
```diff
@@ -176,12 +176,13 @@ export function toMessageSignatureBundle({
 }
 
 function toTransparencyLogEntry(entry: Entry): TransparencyLogEntry {
-  const set = Buffer.from(entry.verification.signedEntryTimestamp, 'base64');
+  const b64SET = entry.verification?.signedEntryTimestamp || '';
+  const set = Buffer.from(b64SET, 'base64');
```
bdehamer, line 180: This was the one type that actually changed -- `verification` is potentially `undefined` so we need to account for a missing `signedEntryTimestamp`. This _should_ never happen, but we'll default to an empty string if it does.

feelepxyz (approved):
Nice!

PR #491, opened 2023-05-16
add toThrowWithCode jest matcher

bdehamer:
#### Summary
Adds a new custom Jest matcher to help in testing functions where we want to check that a custom error was thrown and also that the error has a specific code.

Previously, it was incredibly clunky to test for this condition, but now we can match the error class and code with a single matcher.

```typescript
expect(() => myfunc()).toThrowWithCode(InternalError, 'FOO_ERROR')
```
or
```typescript
expect(myfunc()).rejects.toThrowWithCode(InternalError, 'FOO_ERROR')
```


packages/client/src/__tests__/tsa/index.test.ts
```diff
@@ -63,20 +64,9 @@ describe('TSAClient', () => {
...
-
-      it('has the expected error code', async () => {
-        await subject
-          .createTimestamp(signature)
-          .then(() => {
-            throw Error('expected error');
-          })
-          .catch((err) => {
-            expect(err.code).toEqual('TSA_CREATE_TIMESTAMP_ERROR');
-          });
```
bdehamer, line 79: The helper keeps us from writing tests like this where we have to write a dedicate test case to catch the error and check the code.

PR #500, opened 2023-05-19
Jest Refactoring

bdehamer:
#### Summary
A bunch of random, Jest-related refactoring:

* Adds a basic unit test to the `rekor-types` package just to ensure that we're exporting all of the types we expect.
* Updates the CI GHA workflow to save the Jest cache between runs. This looks to be improving our overall CI time by about 1 minute (20%) when executing the integration tests across the matrix of different node/OS versions.
* Updates the `package.json` to ensure that the project is always built before the tests are run. This works-around an issue with using `jest`/`ts-jest` in a monorepo where you have cross-project references.


.github/workflows/ci.yml
```diff
@@ -59,11 +62,14 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
         cache: npm
+    - name: Manage Jest cache
+      uses: actions/cache@v3
+      with:
+        path: ${{ matrix.platform.jest-cache }}
+        key: ${{ hashFiles('**/package-lock.json') }}
```
bdehamer, line 69: We'll start w/ a fresh cache anytime the `package-lock.json` is updated. This seemed like a reasonable balance between having something cached for most runs . . . but not letting it get totally out-of-date.

feelepxyz (approved):
👌

PR #501, opened 2023-05-20
refactor jest helper into typescript

bdehamer:
#### Summary
Moves the recently add `throwWithCode` Jest matcher from JavaScript (in the `jest.setup.js` file) to a proper TypeScript implementation so we can benefit from all the type checking goodness.


packages/client/jest.setup.js
```diff
@@ -14,62 +14,3 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 console.error = jest.fn();
```
feelepxyz, line 16: Why not change the setup file to typescript? Looks like it's supported here https://jestjs.io/docs/configuration

Also wondering if we want to run the helpers as a setup file before tests run so we don't import it in each file that requires it? e.g. https://jestjs.io/docs/configuration#setupfiles-array

bdehamer: > Why not change the setup file to typescript?

Initially, I was hesitant to do this because I have the TypeScript root for the project set to the "./src" directory and I didn't want to re-configure that to include the "jest.setup.ts" file . . . turns out I can just exclude that file as part of the tsconfig and it seems to work just fine (I still get type-checking on "jest.setup" in my IDE and it seems to be loaded correctly at test-time).

> Also wondering if we want to run the helpers as a setup file before tests run so we don't import it in each file that requires it?

There is some really messy stuff that happens under the covers when you write your Jest tests in TypeScript -- turns out that the test files aren't compiled by the TypeScript compiler which is used for the project source code. Instead, the `ts-jest` plugin uses Babel to transpile the tests and then feed them into Jest. This leads to all sorts of weird behavior when it comes to type resolution. While I can get my custom helper loaded automatically via the "jest.setup" file, I could NOT find any reliable way to get the associated type declarations auto-loaded in a way that worked both when running tests from the terminal AND also within VSCode. The only way I could make everyone happy was to explicitly load the types in the tests where I needed them.

For now, I'm gonna leave the `import ../__helpers__` in the places I need it so that I can get the types imported. At some point, I may even add some actual helper functions in here but, for now, its just a funky way to get my extensions to the Jest global types loaded 🫠

feelepxyz (approved):

packages/client/jest.setup.js
feelepxyz: > I could NOT find any reliable way to get the associated type declarations auto-loaded in a way that worked both when running tests from the terminal AND also within VSCode. The only way I could make everyone happy was to explicitly load the types in the tests where I needed them.

Ah yes this makes sense! Seems like a worthwhile compromise to explicitly load the types 👍  would be super annoying if you didn't get types in the editor.

PR #506, opened 2023-05-22
export IdentityProvider interface

bdehamer:
#### Summary
Working on a plan to move the OAuthIdentityProvider into a separate package (it's really only useful when using the CLI and has no need to be part of the core library), but I need to export this interface so I can consume it in the new package.


feelepxyz (approved):
Makes sense to move the oauth provider to the CLI 💯

PR #515, opened 2023-05-26
move jest extension types to dedicated package

bdehamer:
#### Summary
Fixes the janky work-around I was using to get the proper types loaded for my custom extensions to the Jest `Matchers` interface. I realized that TypeScript will automatically load any types which are shipped in packages in the `@types` namespace (this is how all of the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) stuff works without needing to explicitly import anything). 

By placing my Jest extension types in an `@types`-scoped package (`@types/sigstore-jest-extended` in this case) I can get the TypeScript transpiler (and VSCode) to automatically load my types . . . and since I only need these at dev-time I can throw them in a private package in the monorepo (there's no reason to actually publish this package to npm or add it to the official DefinitelyTyped repo).


packages/jest-types/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@types/sigstore-jest-extended",
+  "private": "true",
```
bdehamer, line 3: This is the magic . . . putting this package in the `@types` namespace grants it special treatment in the TS world and marking it "private" means that it will never by published to npm.

packages/client/src/__tests__/ca/index.test.ts
```diff
@@ -17,7 +17,6 @@ import crypto from 'crypto';
 import nock from 'nock';
 import { CAClient } from '../../ca';
 import { InternalError } from '../../error';
-import '../__helpers__';
```
bdehamer, line 20: No more weird `import`s required 🥳

feelepxyz: 🥳  very nice!

feelepxyz (approved):
🎉

PR #516, opened 2023-05-26
add script to CLI to regenerate readme

bdehamer:
#### Summary
Adds a script to the `package.json` for the CLI package to regenerate and fixup the "README.md" file. There is a bug in the `oclif` readme generator which outputs a badly formed link in the documentation if your CLI project is in a monorepo. I've added a `sed` command to run after the readme generation which strips the bad links from the file.


packages/cli/package.json
```diff
@@ -27,9 +27,9 @@
     "prebuild": "npm run clean",
     "build": "tsc -b",
     "postpack": "shx rm -f oclif.manifest.json",
-    "prepack": "npm run build && oclif manifest && oclif readme",
+    "prepack": "npm run build && oclif manifest && npm run readme",
     "test": "jest",
-    "version": "oclif readme && git add README.md"
+    "readme": "oclif readme && shx sed -i \"s/^_See code:.*$//g\" README.md"
```
feelepxyz, line 32: Worth upstreaming a fix here in case its small?

feelepxyz: Related to this bug or something more specific around monorepos? https://github.com/oclif/oclif/issues/1128

feelepxyz (approved):

packages/cli/package.json
bdehamer: I poked at the code a bit and it wasn't obvious how to resolve. Will log a bug.

PR #521, opened 2023-05-26
consume @sigstore/tuf from sigstore client

bdehamer:
#### Summary
Removes Sigstore TUF client from the `sigstore` package. Uses the version in the new `@sigstore/tuf` package instead.


feelepxyz (approved):
Code change looks good, just wondering if we need to publish `@sigstore/tuf` first?

PR #525, opened 2023-05-28
update generated code to reflect new required fields in intoto v0.0.2

bobcallaway:
This should not have any functional changes, but updates the `calculateDSSEHash` helper to reflect that `sig` and `publicKey` are now required in proposed entry schema.


bdehamer (approved):
👍

Issue #526, opened 2023-05-29
change default rekor type for attestations from 'intoto:0.0.2' to 'dsse'


bobcallaway:
**Description**

This has not been rolled out into the production sigstore environment yet, but it is available in staging (rekor.sigstage.dev).

https://github.com/sigstore/rekor/pull/1487 added support for a new pluggable type `dsse` which has the following benefits:

- The JSON schema more clearly separates the proposed entry from what is persisted to simplify client expectations
- The DSSE envelope is provided as a JSON string, rather than double-encoding values in base64
- The DSSE envelope is not stored by Rekor; this is to set the client expectation that attestation storage in Rekor should not be relied upon for attestation discovery.

bdehamer:
The default was finally changed in the v3.0.0 release of the `@sigstore/sign` library. See [release notes](https://github.com/sigstore/sigstore-js/blob/main/packages/sign/CHANGELOG.md#300).

PR #527, opened 2023-05-29
Adds support for new Rekor 'dsse' entry type

bobcallaway:
#### Summary
This add support for uploading Rekor entries about DSSE envelope-based attestations using the new `dsse` type added in https://github.com/sigstore/rekor/pull/1487. 

This PR builds on https://github.com/sigstore/sigstore-js/pull/525 (given the codegen changes can not be cleanly separated).

#### Release Note
- Adds support for uploading entries to Rekor using `dsse` pluggable type

part of #526


packages/client/src/tlog/verify/body.ts
```diff
@@ -104,6 +132,42 @@ function verifyHashedRekordTLogBody(
...
+  // Turns them into base64-encoded strings for comparison
+  const dsseSigs = dsse.signatures.map((signature) =>
+    signature.sig.toString('base64')
+  );
+
+  // Collect all of the signatures from the tlog entry
+  // Remember that tlog signatures are double base64-encoded
+  const tlogSigs = tlogEntry.spec.signatures?.map((signature) =>
+    signature.signature ? enc.base64Decode(signature.signature) : ''
+  );
```
bdehamer, line 150: Is this comment about the double base64 encoding still true for the DSSE type?

bdehamer (commented):
@bobcallaway I'd like to do try generating/verifying some bundles with the new DSSE Rekor type -- has the build with the DSSE type been deployed to production yet?

Before I merge this, I'm going to add some unit tests around the new verification code as well.

packages/client/src/tlog/verify/body.ts
bobcallaway: do you have docs somewhere on how to target staging with the client? I had to make several code changes when I was doing that.

bdehamer: There's a new CLI in the "packages/cli" directory which supports a flag for overriding the rekor URL. From the root of the project you can do:

```
./packages/cli/bin/dev attest --help
```

That should show you the various flags

Issue #499, opened 2023-05-19
Signing Abstractions


bdehamer:
Toying with some ideas for refactoring the signing workflow in `sigstore-js` . . .

## Considerations

So, you want to use Sigtore to sign a _thing_ . . . I have some questions?

**How do you want the signature to be packaged?**

Do you want a naked signature, or would you like it packaged into a [DSSE envelope](https://github.com/secure-systems-lab/dsse) along with the artifact being signed? The answer to this question obviously influences the returned format of the signature, but also (less intuitively) has implications for how the artifact needs to be prepared before the signature can be generated (see [PAE](https://github.com/secure-systems-lab/dsse) for DSSE).

**What sort of keypair are you going to use to generate the signature?**

Are you providing your own key or using a KMS to do the signing? Do you want to use the Sigstore "keyless" approach with a certificate issued by Fulcio? 

**Which entities do you want to witness the generated signature?**

Are you going to send the signature to the Rekor transparency log and bundle the SET with your signature? Maybe you want to send your signature to multiple Rekor instances? Are you going to generate a signed timestamp and bundle that with your signature? Perhaps you want to use both Rekor and a Timestamp Authority?

### Signatory
Responsible for generating the signature over the provided data. Returns an `Endorsement` which is a signature and the public key which can be used to verify that signature.

```mermaid
classDiagram
    class Signature{
        byte[] signature
        string key
    }
    class Signer
    <<interface>> Signer
    Signer : +sign(data) Signature

    class KeylessSigner {
        FulcioClient fulcio
        IdentityProvider provider
        +sign(data) Signature
    }

    class KMSSigner {
        KMSClient kms
        +sign(data) Signature
    }

    Signer <|-- KeylessSigner : implements
    Signer <|-- KMSSigner : implements
```

### Witness
Some entity which can provide a counter-signature over the `Signatory`'s endorsement.

```mermaid
classDiagram

    class VerificationMaterial {
      TLogEntry[] tlogEntries
      Timestamp[] timestamps
    }

    class Witness
    <<interface>> Witness
    Witness : +testify(signature) VerificationMaterial

    class TSAWitness {
        TimestampAuthorityClient tsa
        +testify(signature) VerificationMaterial
    }

    class RekorWitness {
        TLogClient tlog
        +testify(signature) VerificationMaterial
    }
    Witness <|-- TSAWitness : implements
    Witness <|-- RekorWitness : implements
```

### BundleBuilder
[...]

PR #535, opened 2023-06-02
New package for Sigstore signing logic

bdehamer:
Moves all of the Sigstore signing logic from the `client` package into a new `sign` package. As part of the move, the code has been refactored to align with the abstractions discussed in https://github.com/sigstore/sigstore-js/issues/499. Among other things, this allows for pluggable signers (beyond the included `FulcioSigner`) and configurable combinations of Rekor/Timestamp Authority integrations.

Check the [README](https://github.com/sigstore/sigstore-js/blob/f50f93db3a978cb29af227d83b11f88b75273f6c/packages/sign/README.md#overview) for an overview of the new code organization.

At some point this will be integrated back into the main `sigstore` package and replace the existing signing workflow.


packages/sign/src/__tests__/bundler/dsse.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  } satisfies Signer;
+
+  describe('constructor', () => {
+    it('should create a new instance', () => {
+      const notary = new DSSEBundleBuilder({ signer: signer, witnesses: [] });
+      expect(notary).toBeDefined();
+    });
+  });
+
+  describe('notarize', () => {
```
feelepxyz, line 48:
```suggestion
  describe('sign', () => {
```

packages/sign/src/__tests__/bundler/dsse.test.ts
```diff
@@ -0,0 +1,134 @@
...
+limitations under the License.
+*/
+import assert from 'assert';
+import { DSSEBundleBuilder } from '../../bundler/dsse';
+import { dsse } from '../../util';
+
+import type { Artifact } from '../../bundler';
+import type { Signature, Signer } from '../../signer';
+
+describe('DSSENotary', () => {
```
feelepxyz, line 23:
```suggestion
describe('DSSEBundleBuilder', () => {
```

packages/sign/src/__tests__/bundler/message.test.ts
```diff
@@ -0,0 +1,90 @@
...
+*/
+import { HashAlgorithm } from '@sigstore/protobuf-specs';
+import assert from 'assert';
+import { MessageBundleBuilder } from '../../bundler/message';
+import { crypto } from '../../util';
+
+import type { Artifact } from '../../bundler';
+import type { Signature, Signer } from '../../signer';
+
+describe('MessageNotary', () => {
```
feelepxyz, line 24:
```suggestion
describe('MessageBundleBuilder', () => {
```

packages/sign/src/__tests__/bundler/message.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    it('should create a new instance', () => {
+      const notary = new MessageBundleBuilder({
+        signer: signer,
+        witnesses: [],
+      });
+      expect(notary).toBeDefined();
+    });
+  });
+
+  describe('notarize', () => {
```
feelepxyz, line 52:
```suggestion
  describe('sign', () => {
```

feelepxyz (approved):
Epic refactor! The interfaces make sense to me and feel clearly named. I tried to skim through most files and only spotted some small nits.

packages/sign/src/witness/tlog/index.ts
```diff
@@ -0,0 +1,109 @@
...
+    NonNullable<Entry['verification']>['signedEntryTimestamp']
+  >
+): TransparencyLogEntry['inclusionPromise'] {
+  return {
+    signedEntryTimestamp: Buffer.from(promise, 'base64'),
+  };
+}
+
+function inclusionProof(
+  proof: NonNullable<NonNullable<Entry['verification']>['inclusionProof']>
```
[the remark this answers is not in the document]
bdehamer: The `Entry` type from Rekor gets code-gen'ed as:

```typescript
type Entry {
  verification?: {
    inclusionProof?: InclusionProof | undefined;
  } | undefined;
};
```

For this argument, I just want to the `InclusionProof` type -- it isn't exported from the Rekor package so I have to fish it out of the `Entry` type while working around all of the `undefined` options. Essentially, it boils down to:

> Give me the type of `Entry.verification.inclusionProof` but pretend everything is non-nullable 

Since both Rekor and the Sigstore Bundle define their own (and non-compatible versions) of things like `InclusionProof` and `InclusionPromise` I opted to extract the types from their containing type (either `Entry` or `Bundle`) to provide some context as to which thing we're dealing with. I could export the naked type . . . but then when you see a reference to something like `InclusionProof` it wouldn't be obvious which one you were looking at.

PR #550, opened 2023-06-10
sigstore type refactoring

bdehamer:
#### Summary
Cleans-up some of the internal type definitions. Previously, we were using a couple different flavors of the `Bundle` type:

* The raw `Bundle` generated from the Sigstore protobufs
* A narrowed `BundleWithVerificationMaterial` that we used for the verification flows which ensures that there is non-null verification material included in the bundle.
* A narrowed `ValidBundle` which is guaranteed to contain both verification material and some sort of signature.

Given that we validate any bundles we receive from calling code and we ONLY generate valid bundles internally we can simplify things by using the `ValidBundle` as the only type that we pass around internally. This reduces type clutter and eliminates some unnecessary non-null checks.


packages/client/src/types/sigstore/index.ts
```diff
@@ -13,26 +13,54 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  RFC3161SignedTimestamp,
+  Signature,
+  SubjectAlternativeName,
+  TimestampVerificationData,
+  TransparencyLogEntry,
+  TransparencyLogInstance,
+  TrustedRoot,
+  X509Certificate,
+  X509CertificateChain,
+} from '@sigstore/protobuf-specs';
```
bdehamer, line 61: Instead of exporting EVERYTHING out of `@sigstore/protobuf-specs` I'm trying to be a bit more targeted and export only the types we're actually using. This will help us keep track of what things we're using and help inform what sorts of changes in the upstream package may be breaking.

PR #552, opened 2023-06-12
change default Rekor entry type for intoto attestations to 'dsse'

bobcallaway:
Fixes: #526 

#### Summary
This causes uploads to Rekor for intoto attestations to use the newly-added `dsse` type instead of the `intoto:0.0.2` type. In addition to being simpler to deal with, Rekor will not persist the attestations for `dsse` entries. 

#### Release Note
Uploads to Rekor for intoto attestations now use the `dsse` Rekor type, which will result in the attestation *not* being stored by Rekor.


bdehamer:
@bobcallaway thanks for all the work on the support for the new Rekor type. For the npm use case we still have some work to do on the registry to support the `dsse` type so we won't be merging this until we've got everything ready on the back-end.

bdehamer:
With the recent refactoring of the signing logic, this PR is no longer valid. I'm going to close it in favor of https://github.com/sigstore/sigstore-js/pull/631 which does the same thing (changes the default Rekor type from "intoto" to "dsse")

PR #559, opened 2023-06-13
export new createVerifier function

bdehamer:
#### Summary
Exports a new `createVerifier` function which can be used in scenarios where the client wants to verify a number of bundles in succession without incurring the repeated overhead of initializing the TUF client.

When you call the existing `verify` function it starts by initializing the TUF client and retrieving the "trusted_root.json" target containing all of the Sigstore key material. Even in situations where the local TUF cache is fully populated and up-to-date, this results in HTTP requests to check both the "root.json" and "timestamp.json" from the Sigstore TUF repository. If you're invoking `verify` in a loop (see [`npm/pacote`](https://github.com/npm/pacote/blob/main/lib/registry.js#L302)) this can result in significant overhead.

The new `createVerifier` function performs the initial TUF initialization and then returns a `BundleVerifier` instance with a `verify` method which can be used to verify Sigstore bundles without re-requesting the "trusted_root.json".


packages/client/src/sigstore.ts
```diff
@@ -81,6 +81,23 @@ export async function verify(
   return verifier.verify(deserializedBundle, opts, payload);
 }
 
+export async function basicVerifier(options: config.BasicVerifierOptions) {
```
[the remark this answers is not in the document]
feelepxyz: looks like the issue is with this line: https://github.com/sigstore/sigstore-js/blob/18a4037f2578068cdaee58fe8b3b5d174958cb1f/packages/tuf/src/index.ts#L39

throwing ideas out, could we provide a way to pass in the tuf client as part of verify or something similar? if none are provided, init and refresh one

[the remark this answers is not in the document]
bdehamer: @feelepxyz I refactored this a bit after we talked. The new function is named `createVerifier` and it returns a `BundleVerifier` interface (which itself provides a `verify` function).

.changeset/silly-turtles-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'sigstore': minor
+---
+
+Exports new `basicVerifier` function
```
feelepxyz, line 5:
```suggestion
Exports new `createVerifier` function
```

packages/client/src/sigstore.ts
```diff
@@ -81,6 +81,31 @@ export async function verify(
...
 }
 
+export interface BundleVerifier {
+  verify(bundle: sigstore.SerializedBundle): void;
+}
+
+export async function createVerifier(
+  options: config.BasicVerifierOptions
+): Promise<BundleVerifier> {
+  const trustedRoot = await tuf.getTrustedRoot({
```
feelepxyz, line 91: What are the failure cases for the tuf trust root becoming stale/out of date? Does calling verify trigger a refresh if needed or does it fail?

packages/client/src/sigstore.ts
```diff
@@ -81,6 +81,31 @@ export async function verify(
   return verifier.verify(deserializedBundle, opts, payload);
 }
 
+export interface BundleVerifier {
+  verify(bundle: sigstore.SerializedBundle): void;
+}
+
+export async function createVerifier(
```
feelepxyz, line 88: I like this interface! A async factory function seems like a nice and simple way to handle the initialisation.

packages/client/src/sigstore.ts
bdehamer: The worst that could happen here would be if someone were using the verifier at the same time that Sigstore's "trusted_root.json" file were updated. If that update included the revocation or expiration of a previously valid key we might incorrectly verify a bundle that should be considered invalid.

Unless someone were to instantiate a verifier and use it in some extremely long-running process, I can't see this being a problem. In the npm CLI use case, the TUF data will be refreshed with each no invocation of `npm audit signatures`

feelepxyz (approved):

PR #579, opened 2023-06-30
refactor merkle tree inclusion proof verification

bdehamer:
#### Summary
Clean-up the merkle tree inclusion proof verification logic (see https://github.com/sigstore/protobuf-specs/issues/82) . This was originally implemented before the Sigstore bundle format was even defined and was never integrated into the verification workflow. The change here refactors the merkle verification logic into a form that can be applied to the `TransparenclyLogEntry`s found in the bundle.

This logic is still NOT integrated into the verification workflow (which is why I'm not adding a changeset entry with this PR) but does prepare us for adding this in the future.


packages/client/src/tlog/verify/merkle.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  if (logIndex >= treeSize) {
+    throw new VerificationError('inclusion proof index exceeds size of tree');
+  }
+
+  // Figure out which subset of hashes corresponds to the inner and border
+  // nodes
+  const { inner, border } = decompInclProof(logIndex, treeSize);
+
+  if (inclusionProof.hashes.length != inner + border) {
```
feelepxyz, line 43: should this use string inequality?
```suggestion
  if (inclusionProof.hashes.length !== inner + border) {
```

packages/client/src/tlog/verify/merkle.ts
```diff
@@ -0,0 +1,119 @@
...
+  const inclusionProof = entry.inclusionProof;
+
+  if (!inclusionProof) {
+    throw new VerificationError('tlog entry has no inclusion proof');
+  }
+
+  const logIndex = BigInt(inclusionProof.logIndex);
+  const treeSize = BigInt(inclusionProof.treeSize);
+
+  if (logIndex >= treeSize) {
```
feelepxyz, line 35: should this also check if the logIndex is positive? (I asked copilot to write me a merkle tree verifier and it added this 🤷 )
```suggestion
  if (logIndex < 0n || logIndex >= treeSize) {
```

packages/client/src/tlog/verify/merkle.ts
```diff
@@ -0,0 +1,119 @@
...
+  const leafHash = hashLeaf(entry.canonicalizedBody);
+
+  // Chain the hashes belonging to the inner and border portions
+  const calculatedHash = chainBorderRight(
+    chainInner(leafHash, innerHashes, logIndex),
+    borderHashes
+  );
+
+  // Calculated hash should match the root hash in the inclusion proof
+  return calculatedHash.equals(inclusionProof.rootHash);
```
feelepxyz, line 60: Poking around with copilot.. it's suggesting we rewrite this as a constant time comparison function to prevent timing attacks :) 

googling around, looks like there's a function in node to do this but [throws on mismatch](https://til.simonwillison.net/node/constant-time-compare-strings) so we could wrap it like: 
```
const { timingSafeEqual } = require('crypto');

const compare = (a, b) => {
    try {
        return timingSafeEqual(a, b);
    } catch {
        return false;
    }
};
```

packages/client/src/tlog/verify/merkle.ts
bdehamer: Updated

packages/client/src/tlog/verify/merkle.ts
bdehamer: This is a nice find! I suspect that this is being used internally as part of the crypto signature verification logic, but I'm wondering if there are other places where I should employ this 🤔

feelepxyz (approved):

packages/client/src/tlog/verify/merkle.ts
feelepxyz: > being used internally as part of the crypto signature verification logic

Yeah would hope so!

Maybe worth a quick audit to see if we do any manual comparisons?

PR #581, opened 2023-06-30
fix @types/make-fetch-happen dependency in tuf pkg

bdehamer:
#### Summary
In the `@sigstore/tuf` pacakge, move the `@types/make-fetch-happen` dependency to a dev-only dependency.


feelepxyz (approved):

ejahnGithub (approved):
Lgtm

PR #589, opened 2023-07-03
add func for tlog checkpoint verification

bdehamer:
#### Summary
Adds a function which can perform verification of checkpoint data in tlog entries. 

Will make another PR to integrate this function (and the [inclusion proof verification](https://github.com/sigstore/sigstore-js/pull/579)) into the verification workflow.


packages/client/src/tlog/verify/checkpoint.ts
```diff
@@ -0,0 +1,231 @@
...
+      const sigBytes = Buffer.from(signature, 'base64');
+
+      if (sigBytes.length < 5) {
+        throw new VerificationError('malformed checkpoint: invalid signature');
+      }
+
+      return {
+        name,
+        keyHint: sigBytes.subarray(0, 4),
+        signature: sigBytes.subarray(4),
```
feelepxyz, line 116: Poking at the python implementation this jumped out: https://github.com/sigstore/sigstore-python/blob/bbf0ac83b18dda81aac1e472982507b91e34a8f8/sigstore/_internal/rekor/checkpoint.py#L165

Any idea why they seem to be encoding the sig as base64? Or is it already double encoded or something?

feelepxyz: oh think I found how this works, looks like it's decoded here again: https://github.com/sigstore/sigstore-python/blob/main/sigstore/_internal/rekor/checkpoint.py#L184

packages/client/src/tlog/verify/checkpoint.ts
```diff
@@ -0,0 +1,231 @@
...
+  // corresponding transparency log is looked up by the key hint and the
+  // signature is verified against the public key in the transparency log.
+  // Throws an error if any of the signatures are invalid.
+  public verify(tlogs: ValidTransparencyLogInstance[]): boolean {
+    const data = Buffer.from(this.note, 'utf-8');
+
+    return this.signatures.every((signature) => {
+      // Find the transparency log instance with the matching key hint
+      const tlog = tlogs.find((tlog) =>
+        crypto.bufferEqual(tlog.logId.keyId.subarray(0, 4), signature.keyHint)
```
feelepxyz, line 137: potentially nit pick: should we recalculate the key id by hashing the public key? looks like they do this in sigstore-python but can't tell if this is overkill? https://github.com/sigstore/sigstore-python/blob/main/sigstore/_internal/rekor/checkpoint.py#L223

feelepxyz (approved):
This looks good to me, had a minor comment about potentially rehashing the public key id/hint?

PR #601, opened 2023-07-10
integrate @sigstore/bundle into client

bdehamer:
#### Summary
Integrates the new `@sigstore/bundle` package into the `client` package. This PR is mostly just changes to import statements and deletion of the code which now lives in the `bundle` package.


packages/client/src/__tests__/tlog/verify/set.test.ts
```diff
@@ -101,6 +101,23 @@ describe('verifyTLogSET', () => {
...
+              validFor: { start: undefined },
+            },
+          },
+        ];
+
+        it('returns false', () => {
+          expect(verifyTLogSET(entry, tlogs)).toBe(false);
+        });
+      });
+
```
bdehamer, line 120: Covers a scenario (a key in the `trusted_root.json` which was missing the actual bytes of the key) which was not previously tested.

packages/client/src/tlog/verify/checkpoint.ts
```diff
@@ -196,36 +188,38 @@ class LogCheckpoint {
...
-): ValidTransparencyLogInstance[] {
+): sigstore.ViableTransparencyLogInstance[] {
   const targetDate = new Date(Number(integratedTime) * 1000);
-  return tlogInstances.filter((tlog): tlog is ValidTransparencyLogInstance => {
-    // Must have a log ID
-    if (!tlog.logId) {
-      return false;
-    }
+  return tlogInstances.filter(
+    (tlog): tlog is sigstore.ViableTransparencyLogInstance => {
```
bdehamer, line 194: This is a neat trick . . . specifying a type assertion in the return value of the anonymous function called by `Array.filter` allows us to narrow the type of the return value.  We call `filter` on an array of `TransparencyLogInstance`s but the returned array contains `ViableTransparencyLogInstance`s (i.e. instances where we've validated that the required log ID and public key are present)

packages/client/src/tlog/verify/index.ts
```diff
@@ -50,16 +56,11 @@ export function verifyTLogEntries(
...
+  entry: TransparencyLogEntry,
+  bundleContent: Bundle['content'],
   tlogs: sigstore.TransparencyLogInstance[],
   signingCert?: x509Certificate
 ): boolean {
-  // Check that the TLog entry has the fields necessary for verification
-  if (!sigstore.isVerifiableTransparencyLogEntry(entry)) {
-    return false;
-  }
-
```
bdehamer, line 62: This check is now happening as part of the bundle validation process.

packages/client/src/tlog/verify/set.ts
```diff
@@ -41,10 +43,6 @@ export function verifyTLogSET(
 
   // Check to see if we can verify the SET against any of the valid tlogs
   return validTLogs.some((tlog) => {
-    if (!tlog.publicKey?.rawBytes) {
-      return false;
-    }
-
```
bdehamer, line 47: This check happens as part of the filter function above now

packages/client/package.json
```diff
@@ -36,6 +36,7 @@
     "@types/make-fetch-happen": "^10.0.0"
   },
   "dependencies": {
+    "@sigstore/bundle": "^0.0.0",
```
feelepxyz, line 39: Do we not want to publish an actual version of this package? Or follow up once we publish sigstore-js?

bdehamer: Yeah, I'll be bumping to 1.0 before I cut a release with any of this stuff

PR #611, opened 2023-07-12
backward compat verification tests

bdehamer:
#### Summary
Adds a new GHA check to test backward compatibility for verifying different bundle versions and rekor entry types.


feelepxyz (approved):
Nice!

PR #628, opened 2023-07-20
clean-up for @sigstore/sign package

bdehamer:
#### Summary
Renames the `MessageBundleBuilder` class to `MessageSignatureBundleBuilder`. Also fixes some incorrect test descriptions and narrows types in a few places.


packages/sign/src/bundler/base.ts
```diff
@@ -43,7 +43,9 @@ export interface BundleBuilder {
 // provides a the basic wokflow for signing and witnessing an artifact.
 // Subclasses must implement the `package` method to assemble a valid bundle
 // with the generated signature and verification material.
-export abstract class BaseBundleBuilder implements BundleBuilder {
+export abstract class BaseBundleBuilder<T extends Bundle>
+  implements BundleBuilder
+{
```
bdehamer, line 48: Using a generic here allows sub-classes to return narrowed-versions of the `Bundle` type (like `BundleWithDSSEEnvelope` and `BundleWithMessageSignature`) which can save consumers from having to do a bunch of type checking on the return value.

feelepxyz (approved):

PR #629, opened 2023-07-20
integrate @sigstore/sign into client package

bdehamer:
#### Summary
Depends on https://github.com/sigstore/sigstore-js/pull/628

Integrates the new `@sigstore/sign` package into the main `sigstore` package. This resulted in some minor refactoring internal to `sigstore` and the deletion of A LOT of code. Does NOT change the public interface of the `sigstore` client.


packages/client/src/sigstore-utils.ts
```diff
@@ -30,45 +39,50 @@ export async function createDSSEEnvelope(
...
-    ],
-  };
-
-  return sigstore.Envelope.toJSON(envelope) as SerializedEnvelope;
+  const bundler = createBundleBuilder('dsseEnvelope', {
+    signer: options.signer,
+    tlogUpload: false,
+  });
+  const bundle = await bundler.create({ data: payload, type: payloadType });
+  return envelopeToJSON(bundle.content.dsseEnvelope);
```
bdehamer, line 47: Note how the `createDSSEEnvelope` and `createRekorEntry` functions we export here (exclusively for use by the npm registry) are now implemented by calls to the `DSSEBundleBuilder` and the `RekorWitness` (both exported from `@sigstore/sign`). 

In the future, we can probably just move these functions into the places in the registry where they are needed.

packages/client/src/types/signature.ts
```diff
@@ -32,18 +33,45 @@ export type SignatureMaterial = {
...
-    },
-    certificates: undefined,
-  };
+type CallbackSignerOptions = {
+  signer: SignerFunc;
+};
+
+// Adapter to allow the legacy SignerFunc callback to be used as a new Signer
+// interface.
+export class CallbackSigner implements Signer {
```
feelepxyz, line 42: Should this class live in the types directory? Wondering if it will get lost here

packages/client/src/sigstore-utils.ts
feelepxyz: Nice! Def up for migrating this to the registry and adding these utils the deprecations for 2.0 🙌

packages/client/src/sigstore-utils.ts
```diff
@@ -30,45 +39,50 @@ export async function createDSSEEnvelope(
...
-  const bundle = sigstore.toDSSEBundle({
-    envelope,
-    signature: sigMaterial,
-    tlogEntry: entry,
-  });
+  const vm = await tlog.testify(content, publicKey);
+
+  const bundle: Bundle = {
+    mediaType: BUNDLE_V01_MEDIA_TYPE,
+    content,
```
feelepxyz, line 76: I think this will result in something like this:

```
mediaType: ...
content: {
  $case: 'dsseEnvelope',
  dsseEnvelope: envelope,
}
``` 

Do we also need to omit the `$case` key or is this handled by `bundleToJSON`:
```suggestion
    dsseEnvelope: content.dsseEnvelope,
```

Seems worth adding some tests around this to match against the expected bundle output.

feelepxyz: or wait, is this all handled by `bundleToJSON`?

packages/client/src/config.ts
```diff
@@ -78,32 +74,118 @@ export const DEFAULT_REKOR_URL = 'https://rekor.sigstore.dev';
...
+  }
 }
 
-export function createTSAClient(
-  options: TSAOptions & FetchOptions
-): TSA | undefined {
-  return options.tsaServerURL
-    ? new TSAClient({
+// Instantiate a collection of witnesses based on the supplied options.
+function initWitnesses(options: SignOptions): Witness[] {
```
feelepxyz, line 143: Neat!

packages/client/src/sigstore-utils.ts
feelepxyz: I think I'm just getting tripped up by the manual bundle construction here. Would it be worth adding some helpers to the `bundle` package? Wondering later down the line if it would be easier to have everything bundle related in one place so we can do roll out bigger changes and not miss things elsewhere.

feelepxyz (approved):
Very nice 👍  some minor comments that don't need to block this.

PR #631, opened 2023-07-21
use 'dsse' as the default rekor type when signing

### LINKED ISSUE #526: change default rekor type for attestations from 'intoto:0.0.2' to 'dsse'

bdehamer:
Fixes #526 
Replaces https://github.com/sigstore/sigstore-js/pull/552

#### Summary
Updates `@sigstore/sign` to use "dsse" as the default Rekor type when submitting DSSE-wrapped payloads. This replaces the current "intoto" type currently in use.

NOTE: Do not merge until the necessary support has been added to the rest of the stack accepting npm provenance statements.

### DIFF
--- a/.changeset/blue-weeks-hope.md
+++ b/.changeset/blue-weeks-hope.md
@@ -1,0 +1,5 @@
+---
+'@sigstore/sign': minor
+---
+
+Use `dsse` as the default Rekor entry type (instead of `intoto`) when submitting DSSE payloads
--- a/packages/sign/src/__tests__/integration.test.ts
+++ b/packages/sign/src/__tests__/integration.test.ts
@@ -90,7 +90,7 @@ describe('artifact signing', () => {
 
       expect(bundle.verificationMaterial.tlogEntries).toHaveLength(1);
       expect(bundle.verificationMaterial.tlogEntries[0].kindVersion.kind).toBe(
-        'intoto'
+        'dsse'
       );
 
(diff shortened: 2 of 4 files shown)


haydentherapper:
Hey! Just wanted to check if you know when this can be merged.

bdehamer:
@haydentherapper I don't think we're quite ready to make the `dsse` Rekor type the default just yet -- this has implications for some of the older versions of `npm` which have versions of this library bundled (don't want to get in a position where older clients aren't able to verify newly signed packages). I'm not sure what the exact trigger will be, but when the usage numbers for some of those older `npm` versions start to dip we will consider switching the default.

That said, GH is about to deploy a new usage of this library where we will be using the `dsse` Rekor type exclusively.

bdehamer:
Closing in favor of https://github.com/sigstore/sigstore-js/pull/1270

PR #640, opened 2023-07-26
update mock with tlog inclusion proof

bdehamer:
#### Summary
Updates the `@sigstore/mock` package to add a verifiable inclusion proof to any response from the mock TLog. To limit the complexity of calculating the inclusion proof, we pretend that every new entry added to the log is the first entry in the tree.


feelepxyz (approved):
Very cool!

PR #646, opened 2023-07-27
fix bug in merkle tree verification

bdehamer:
#### Summary
Fixes a bug that would result in certain valid inclusion proofs failing verification.

In scenarios where all of the hashes in the inclusion proof were on the right-side of the tree the division of the hashes between the inner- and right-side hashes would be off by one. This stemmed from an incorrect calculation related to the bit length of a number -- when calculating the number of bits required to represent the number `0` the result was `1` when the appropriate answer in this case is `0` (see the implementation for Python's [bit_length()](https://python-reference.readthedocs.io/en/latest/docs/ints/bit_length.html)).


packages/client/src/__tests__/tlog/verify/merkle.test.ts
```diff
@@ -181,4 +181,85 @@ describe('verifyMerkleInclusion', () => {
       );
     });
   });
+
+  describe('when all of the hashes are on the right side of the tree', () => {
```
bdehamer, line 185: Adds a test case which specifically detects this scenario and will fail if the bit-length calculation is wrong.

feelepxyz (approved):

PR #649, opened 2023-07-27
Fix bug in mocked inclusion proof

bdehamer:
#### Summary
Fixes a bug in the inclusion proof returned as part of the mocked transaction log entry. The previous implementation was predicated on a bug that was recently discovered in the inclusion proof verification logic (see #646). With the verification bug fixed, we can need to adjust the fake inclusion proof generated for the mocked Rekor service.

Also took this opportunity to refactor the code in the mock `TLog` class to make SET and inclusion proof calculations more readable.


packages/mock/src/rekor/tlog.ts
```diff
@@ -104,25 +139,12 @@ class TLogImpl implements TLog {
...
     const checkpoint = [note, sigLine, ''].join('\n');
 
-    const proof: InclusionProof = {
-      logIndex: logIndex,
+    return {
+      logIndex: 0,
       treeSize: treeSize,
       checkpoint,
-      hashes: [leafHash.toString('hex')],
+      hashes: [],
```
feelepxyz, line 146: Do you not need any hashes here?

feelepxyz (approved):

packages/mock/src/rekor/tlog.ts
bdehamer: @feelepxyz as a result of fixing that verification bug we can now represent the simplest possible merkle tree which has a single node. In this scenario, the root hash of the tree is equal to the leaf hash of the sole node -- no other node hashes are required to verify the integrity of the tree.

PR #651, opened 2023-07-27
sigstore-in-a-box

bdehamer:
#### Summary
Adds a new (private) `server` project which allows for running an HTTP server that provides mocked versions of all the core Sigstore services.


packages/server/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@sigstore/server",
```
feelepxyz, line 2: nit but should this be called something like `@sigstore/mock-server` to make it clearer it's only for testing purposes?

packages/server/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "version": "0.0.0",
+  "private": "true",
+  "description": "Sigstore Server",
```
feelepxyz, line 5: maybe call out it's only for testing in case we publish this at some point?
```suggestion
  "description": "Sigstore mock server for testing purposes only",
```

feelepxyz (approved):
LGTM! Just wondering if we want to rename this to something like `mock-server`

bdehamer:
> Just wondering if we want to rename this to something like mock-server
I started with that name and then talked myself out of it. But I think you're right, better to be super explicit that this is not for production use. Gonna merge and come back with a separate rename PR

PR #654, opened 2023-07-28
fix bug in mock tlog SET calculation

bdehamer:
#### Summary
Fix bug in `@sigstore/mock` related to the SET calculation for mocked transparency log entries


ejahnGithub (approved):
Lg

Issue #523, opened 2023-05-26
2.0 Release Planning


bdehamer:
**Description**
Based on the feedback we've received and the lessons we've learned from integrating this library into the npm CLI and registry, we have a number of big changes we'd like to make. We're currently working toward a 2.0 release of the main `sigstore` client which will contain breaking changes to the public interface:

* Removal of the basic CLI shipped as part of the `sigstore` package. This CLI was only ever intended to be a demo/testing tool and has been superseded by the far more robust CLI now found in the [`@sigstore/cli`](https://www.npmjs.com/package/@sigstore/cli) package.
* Removal of the TUF-related `tuf.getTarget` and `tuf.client` functions from the public interface of the `sigstore` package. All of the functions related to interaction with the Sigstore TUF repository have been moved to the [`@sigstore/tuf`](https://github.com/sigstore/sigstore-js/tree/main/packages/tuf) package.
* Removal of the following OIDC-related options from the public `sign` and `attest` functions:
  * `oidcIssuer`
  * `oidcClientID`
  * `oidcClientSecret`
  * `oidcRedirectURL`
  
  These options all relate to the OAuth identity provider used to retrieve an OIDC token through interactive user authentication. The code supporting interactive authentication has been moved to the CLI ([`@sigstore/cli`](https://www.npmjs.com/package/@sigstore/cli)) and will be removed from the main `sigstore` package. Each of these options is exposed as a command line flag on the CLI's [`attest`](https://github.com/sigstore/sigstore-js/tree/main/packages/cli#sigstore-attest-file) command.

* Removal of the `signer` option from the public `sign` and `attest` functions. The bring-your-own-signer use case will be supported through some new interfaces.
* Removal of the exported `utils` object. This was never part of the documented public interface for the package and mainly serves internal npm use cases. The functions attached to this object will be moved to a separate package.
* Drop support for node 14 (required for inclusion in npm v10)
* Update `make-fetch-happen` to v12 (required for inclusion in npm v10)

bdehamer:
v2 released!

PR #684, opened 2023-08-11
remove sigstore-utils from public interface

bdehamer:
#### Summary
Removes the `sigstore-utils` object from the public interface of the `sigstore` package. This was never considered part of the supported interface of the package and is now being removed entirely.

See https://github.com/sigstore/sigstore-js/issues/523


feelepxyz (dismissed):
🎉

feelepxyz (approved):

PR #686, opened 2023-08-11
remove custom signer option from client package

bdehamer:
#### Summary
Removes `signer` as an option from the `sign` and `attest` functions. Use cases which require something other than Fulcio-style keyless signing can use the `@sigstore/sign` package directly and provide a custom implementation for the `Signer` interface.

Branches from #684 

See #523


feelepxyz (approved):
🎉

PR #694, opened 2023-08-16
Clean-up public interface of `sigstore` package

bdehamer:
#### Summary
The primary change here is a flattening of the exports from the `sigstore` package. Previously this package exported a single `sigstore` object under which contained the various functions/errors/types. With this change, everything that was previously under the `sigstore` object is now exported as a top-level item. This will make it easier for consumers to import only the items they need and also enables type-only imports via `import type`.

The other big change is the inclusion of function overloads for `verify` to better handle use cases where the original artifact does not need to be provided for verification.


feelepxyz (approved):
🎉

Issue #561, opened 2023-06-15
Improving Error Messages, specifically for signing certificate and tlog entry


enteraga6:
**Describe the issue**
When running a workflow that utilizes sigstore-js to generate a token I received the error messages: "error creating signing certificate" and "error creating tlog entry". Running the workflow again after the first two failures resulted in it generating the token successfully, but I am still confused as to the exact reason why it failed the first two times. To help users identify the source of their problem, I propose improving error messages to provide more detail about the specific issue or failure that occurred. This will give users a better understanding of what went wrong and why, which will make it easier for them to troubleshoot the problem and understand why it occurred.

**Relevant Links**
[Code for "error creating signing certificate"](https://github.com/sigstore/sigstore-js/blob/8db6c5c953478eddb94c40f3df418f46c211f125/packages/client/src/ca/index.ts#L70)
[Code for "error creating tlog entry"](https://github.com/sigstore/sigstore-js/blob/8db6c5c953478eddb94c40f3df418f46c211f125/packages/client/src/tlog/index.ts#L100)

**Context if Needed**
These were my workflow runs:
[Run 1: "error creating tlog entry"
](https://github.com/enteraga6/slsa-lvl3-generic-provenance-with-bazel-example/actions/runs/5272925971/jobs/9535768491)[Run 2: "error creating signing certificate" ](https://github.com/enteraga6/slsa-lvl3-generic-provenance-with-bazel-example/actions/runs/5272937339)

bdehamer:
With the 2.1 release, the specific error text returned by Fulcio/Rekor will now be included in the message of the thrown error.

PR #711, opened 2023-08-21
surface error msgs from sigstore services

bdehamer:
Resovles #561 

#### Summary
Improved messaging when throwing errors originating from Sigstore services.  Previously, when responding to an error from Fulcio, the consumer would see a generic errror message like this:

```
InternalError: error creating signing certificate
```

This explains that there was some sort of issue with the creation of the certificate, but doesn't relay the specific issue reported by Fulcio.

With this change, the consumer will see the following:

```
InternalError: error creating signing certificate - (400) There was an error processing the identity token
```

Now they have the HTTP status code and the specific Fulcio-reported error message available.

### DIFF
--- a/packages/sign/src/__tests__/external/error.test.ts
+++ b/packages/sign/src/__tests__/external/error.test.ts
@@ -28,8 +28,8 @@ describe('checkStatus', () => {
       ok: true,
     });
 
-    it('returns the response', () => {
-      expect(checkStatus(response)).toEqual<Response>(response);
+    it('returns the response', async () => {
+      await expect(checkStatus(response)).resolves.toEqual<Response>(response);
     });
   });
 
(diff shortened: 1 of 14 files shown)


packages/sign/src/__tests__/external/error.test.ts
```diff
@@ -40,13 +40,13 @@ describe('checkStatus', () => {
...
-    it('throws an error', () => {
+    it('throws an error', async () => {
       expect.assertions(2);
       try {
-        checkStatus(response);
+        await checkStatus(response);
       } catch (e) {
         assert(e instanceof HTTPError);
-        expect(e.message).toEqual('HTTP Error: 404 Not Found');
+        expect(e.message).toEqual('(404) Not Found');
```
feelepxyz, line 49: Do we want any tests that check the entire error? Looks like these tests are only checking the checkStatus/http error and not the full internal error like: `error creating signing certificate - (400) There was an error processing the identity token`

bdehamer: Good call. Added more tests.

feelepxyz (approved):

Issue #728, opened 2023-08-29
Documentation example for @sigstore/sign does not run


amithkk:
**Description**

The usage example at https://github.com/sigstore/sigstore-js/tree/main/packages/sign has the following snippet:

```
// Sign a thing
const artifact = {
  data: Buffer.from('something to be signed'),
};
const bundle = await bundler.create(artifact);
```

However, this will not work as the example sets up a DSSEBundleBuilder which requires the `type` key to be set on the artifact. Perhaps this example will need to be updated to have `type: 'dsse_envelope'` added to the artifact

**Version**

2.0.0 from NPM

PR #729, opened 2023-08-29
Add 'type' to artifact to align example with docs

amithkk:
Fix for #728


packages/sign/README.md
```diff
@@ -169,6 +169,7 @@ const bundler = new DSSEBundleBuilder({
 
 // Sign a thing
 const artifact = {
+  type: 'dsse_envelope',
```
bdehamer, line 172: Let's update to make it clear that it should be some sort of content type
```suggestion
  type: 'text/plain',
```

bdehamer (changes requested):
Thanks for the fix. Had one suggestion for improvement.

packages/sign/README.md
amithkk: Sounds good

bdehamer (approved):

PR #737, opened 2023-09-01
fix bug in mocked inclusion proof calculation

bdehamer:
#### Summary
Fixes a bug in the `mock` library which was generating an invalid merkle tree inclusion proof.

When calculating the leaf hash for the merkle tree inclusion proof, the base-64 encoding of the tlog entry was being used instead of the raw bytes.

### DIFF
--- a/.changeset/slow-weeks-shop.md
+++ b/.changeset/slow-weeks-shop.md
@@ -1,0 +1,5 @@
+---
+'@sigstore/mock': patch
+---
+
+Fix bug in mocked inclusion proof calculation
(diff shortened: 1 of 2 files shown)


bdehamer:
It's worth noting that there was no test which caught this bug (and no new test added as part of this change to catch future regressions). Fully testing the crypto material generated in this `mock` library would require us to copy-pasta a huge chunk of the verification logic in the `sigstore` package (or create a circular dependency by importing the required functions).

Instead, I plan to add a new case to the existing smoke test which spins-up the mock server and then uses the CLI to sign/verify against the mock. That should surface any future issues like this. First, I need to merge #751 which will allow us to use the CLI to interact w/ arbitrary Sigstore instances.

feelepxyz (approved):

PR #738, opened 2023-09-01
configurable keys for mock services

bdehamer:
#### Summary
Updates the `mockFulcio`, `mockRekor` and `mockTSA` functions to accept an optional keypair. When specified, this keypair will be used as the key for generating certificates (Fulcio and TSA) or generating inclusion proofs/promises (Rekor).

By default, each service generates an ephemeral key that lives only as long as the service instance is in scope. Passing an explicit keypair argument allows the same keys to be used across invocations.


feelepxyz (approved):
Nice one, lgtm!

PR #751, opened 2023-09-07
CLI command to initialize local TUF cache

bdehamer:
#### Summary
Adds a new `initialize` command the CLI which can be used to initialize the local TUF cache (inspired by a [similar command](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/initialize.go) in the cosign CLI). 

Previously, the CLI would only work with the public-good instance of the Sigstore services. This new command allows users to bootstrap their local TUF cache with a custom TUF URL and trusted root, enabling the use of alternative/private Sigstore instances.

Also includes a new option for the `@sigstore/tuf` library which enables forced re-initialization of local TUF cache even if it already exists (allowing users to reset their TUF cache to work with a new Sigstore instance).


packages/cli/README.md
```diff
@@ -80,6 +81,31 @@ DESCRIPTION
...
+FLAGS
+  --force           force initialization even if the cache already exists
+  --mirror=<value>  [default: https://tuf-repo-cdn.sigstore.dev] URL to the Sigstore TUF repository
+  --root=<value>    path to the initial trusted root. Defaults to the embedded root.
+
+DESCRIPTION
+  initialize the Sigstore TUF root to retrieve trusted certificates and keys for verification
+
+ALIASES
+  $ sigstore init
```
feelepxyz, line 101: You trying to remove this with the --no-aliases flag below?

feelepxyz (approved):
This is great, I've wanted this on a few occasions in the past and ended up nuking the folder to force re-fresh 💯

PR #752, opened 2023-09-07
smoke test against mock service stack

bdehamer:
#### Summary
Updates the smoketest workflow to include a test which will sign and verify an artifact using the mock sigstore services.


.github/workflows/smoke-test.yml
```diff
@@ -35,14 +35,67 @@ jobs:
...
+      run: |
+        echo -n "hello world" > artifact
+    - name: Sign artifact
+      run: |
+        ./packages/cli/bin/run attest \
+          --fulcio-url ${SIGSTORE_URL} \
+          --rekor-url ${SIGSTORE_URL} \
+          --tsa-server-url ${SIGSTORE_URL} \
+          --type "text/plain" \
+          --out bundle.json \
```
feelepxyz, line 90: any reason this isn't `bundle.sigstore`?

feelepxyz, line 90: any reason this isn't `bundle.sigstore`?

PR #764, opened 2023-09-15
add issuer extension to mock certs

bdehamer:
#### Summary
Updates the certificates issued by the mock Fulcio service to include the "issuer" extension. If available, the issuer extension is populated from the `iss` claim in the OIDC token; otherwise, a default value is populated.


packages/mock/src/fulcio/handler.ts
```diff
@@ -22,6 +22,9 @@ import type { CA } from './ca';
 
 const CREATE_SIGNING_CERT_PATH = '/api/v2/signingCert';
 const DEFAULT_SUBJECT = 'NO-SUBJECT';
+const DEFAULT_ISSUER = 'https://token.actions.githubusercontent.com';
```
feelepxyz, line 25: Thoughts on setting this to some obviously fake default value? Feels a bit weird having github.com's issuer in the mock :)

bdehamer: Good call.

feelepxyz (approved):

PR #820, opened 2023-10-20
expose rekor option to choose 'intoto' or 'dsse'

bdehamer:
#### Summary
Exposes an option on the `RekorWitness` class which allows the caller to choose between "intoto" and "dsse" entry types when creating a new TLog entry. Will continue to default to "intoto" if no value is specified.


ejahnGithub (approved):
LGTM!

PR #889, opened 2023-12-07
integrate verify package with client

bdehamer:
Integrates the new `@sigstore/verify` package with the `sigstore` client package.

Removes all of the verification code which was copied (#886) into the `@sigstore/verify` package. Also refactors the remaining unit tests to ensure we have 100% code coverage.


ejahnGithub (approved):
look good!

PR #891, opened 2023-12-07
add identity/issuer policy support to verifier

bdehamer:
Updates the verifier with support for verifying the identity contained in the signing certificate. Given that not all Sigstore bundles contain a signing certificate, the `policy` argument containing the expected identity is marked as optional.


ejahnGithub (approved):
LG!

PR #893, opened 2023-12-08
add verification policy support to client package

bdehamer:
Updates the `sigstore` package to translate the input verification options into a `VerificationPolicy` which will be checked when the `Verifier` is invoked.

Includes a refactor of the `verifyPolicy` logic to make it more extensible if we want to support additional certificate extensions beyond the `issuer` in the future.


ejahnGithub (approved):
looks good!

PR #894, opened 2023-12-08
Support for `--trusted-root` flag in conformance tests

bdehamer:
Updates the conformance test CLI to support the new `--trusted-root` flag (see https://github.com/sigstore/sigstore-conformance/pull/101)


feelepxyz (approved):
Nice to see some minor tweaks to the conformance lib and leveraging the different sub-packages 💯

PR #898, opened 2023-12-13
Add most verify options to `sigstore verify` CLI subcommand

AshCorr:
#### Summary

When trying to test a bundle I noticed the CLI was lacking a lot of the `verify` functionality which made it difficult to use!

This PR adds more flags to the `sigstore verify` subcommand to let users access most of the `verify` functionality that the sigsign-js library has.

Looks like this was discussed at one point in https://github.com/sigstore/sigstore-js/pull/477#discussion_r1193873108 and that ideally we'd refactor some of these flags in future, but for now I'm just focusing on getting the functionality working.

#### Release Note

- Added following flags to `sigstore verify` subcommand (most are options from VerifyOptions)
  - `certificate-issuer`
  - `certificate-identity-email`
  - `certificate-identity-uri`
  - `tuf-mirror-url`
  - `tuf-root-path`
  - `tuf-cache-path`
  - `blob-file` File containing data to verify, only required if bundle was not signed using attest
  - `blob` Base64 encoded data to be verified using bundle. Mutually exclusive with `blob-file`


packages/cli/src/commands/verify.ts
```diff
@@ -21,6 +21,47 @@ export default class Verify extends Command {
       min: 0,
       default: 1,
     }),
+    'certificate-issuer': Flags.string({
+      description:
+        "Value that must appear in the signing certificate's issuer extension (OID 1.3.6.1.4.1.57264.1.1). Not verified if no value is supplied",
```
bdehamer, line 26:
```suggestion
        "Value that must appear in the signing certificate's issuer extension (OID 1.3.6.1.4.1.57264.1.1 or 1.3.6.1.4.1.57264.1.8). Not verified if no value is supplied",
```

packages/cli/src/commands/verify.ts
```diff
@@ -21,6 +21,47 @@ export default class Verify extends Command {
...
+    'certificate-identity-uri': Flags.string({
+      description:
+        "URI which must appear in the signing certificate's Subject Alternative Name (SAN) extension. Not verified if no value is supplied",
+      dependsOn: ['certificate-issuer'],
+      exclusive: ['certificate-identity-email'],
+    }),
+    'certificate-oids': Flags.string({
+      description:
+        "A collection of OID/value pairs which must be present in the certificate's extension list. Not verified if no value is supplied",
+    }),
```
bdehamer, line 43: Let's ditch this flag for now. I noticed that it wasn't passed-in to the verifier and it's not implemented yet even if we were to pass it in.

bdehamer (commented):
@AshCorr thanks for the contribution! A couple of minor requests, but this looks great otherwise.

packages/cli/src/commands/verify.ts
AshCorr: Ah yea, forgot about this one! I'll remove it!

bdehamer (changes requested):

packages/cli/src/commands/verify.ts
```diff
@@ -21,6 +21,43 @@ export default class Verify extends Command {
       min: 0,
       default: 1,
     }),
+    'certificate-issuer': Flags.string({
+      description:
+        "Value that must appear in the signing certificate's issuer extension (OID 1.3.6.1.4.1.57264.1.1). Not verified if no value is supplied",
```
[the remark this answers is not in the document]
AshCorr: 🤦

This is why you don't force push, oops

bdehamer (approved):

PR #901, opened 2023-12-15
extract issuer from fulcio cert

bdehamer:
Updates the Fulcio certificate verification logic to account for the fact that there are two different extensions which may contain the name of the certificate issuer. Older instances of Fulcio would encode the issuer in an extension with OID `1.3.6.1.4.1.57264.1.1` while new versions of Fulcio use `1.3.6.1.4.1.57264.1.8`.

To properly verify the certificate issuer, we need to check both extensions to see which one contains the issuer string.


feelepxyz (approved):

ejahnGithub (approved):
lg

PR #908, opened 2023-12-19
Fixes for TSA-generated timestamps in mock package

bdehamer:
Updates to the `@sigstore/mock` package so that the mock TSA instance can issue verifiable timestamps. 

The timestamps which were generated by the mock TSA previously were missing some critical fields which made them unverifiable. The signed timestamp now includes the necessary information about the certificate which was used to do the signing. This is used in the verification process to look-up the correct cert.

This change also introduces a new intermediate certificate with the "Time Stamping" key usage extension which is necessary for generating a valid timestamp (the verification process checks that the issuing certificate has this usage listed).

With these changes I was able to use `openssl ts -verify` to successfully verify a timestamp generated by the mock TSA instance.

### DIFF
--- a/.changeset/nine-donuts-marry.md
+++ b/.changeset/nine-donuts-marry.md
@@ -1,0 +1,5 @@
+---
+"@sigstore/mock": patch
+---
+
+Introduce intermediate certificate for issuing RFC3161 timestamps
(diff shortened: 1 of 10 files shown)


ejahnGithub (approved):
lgtm

PR #909, opened 2023-12-20
Refactor test fixtures for verify package

bdehamer:
Clean-up of the test fixtures used in the `@sigstore/verify` package. Consolidates all of the valid bundle fixtures into a single module and deletes a bunch of un-used fixtures.


ejahnGithub (approved):
look good to me

PR #913, opened 2023-12-21
support for parsing RFC3161 timestamps

bdehamer:
Adds an `RFC3161Timestamp` class to the `@sigstore/core` package. This sets the stage for adding full signed timestamp verification to the `@sigstore/verify` package.

### DIFF
--- a/packages/core/src/__tests__/rfc3161/timestamp.test.ts
+++ b/packages/core/src/__tests__/rfc3161/timestamp.test.ts
@@ -1,0 +1,158 @@
+/*
+Copyright 2023 The Sigstore Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


packages/core/src/__tests__/rfc3161/timestamp.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      const subject = RFC3161Timestamp.parse(ts);
+      const key = createPublicKey(publicKey);
+      const data = artifact;
+
+      it('does not throw an error', () => {
+        expect(() => subject.verify(data, key)).not.toThrow();
+      });
+    });
+
+    describe('when the artifact does NOT the one which was signed', () => {
```
feelepxyz, line 86:
```suggestion
    describe('when the artifact does NOT match the one which was signed', () => {
```

packages/core/src/__tests__/rfc3161/tstinfo.test.ts
```diff
@@ -0,0 +1,71 @@
...
+
+      it('does not throw an error', () => {
+        expect(() => subject.verify(artifact)).not.toThrow();
+      });
+    });
+
+    describe('when the messageImprintHashedMessage does NOT match the artifact', () => {
+      const artifact = Buffer.from('oops');
+
+      it('does not throw an error', () => {
```
feelepxyz, line 64:
```suggestion
      it('throws an error', () => {
```

feelepxyz (approved):
Nice one! Added some minor comments otherwise LGTM 👍

PR #920, opened 2023-12-29
default for isCA value in basic constraints extension

bdehamer:
Updates the `X509BasicConstraintsExtension` in the `@sigstore/core` package to ensure that that `isCA` property defaults to `false` in scenarios where there is no explicit value encoded in the extension (per the [spec](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9)).


ejahnGithub (approved):
lg

PR #922, opened 2023-12-29
expose signature on SignatureContent

bdehamer:
Updates the `SignatureContent` type to expose a public `signature` property which gives direct access to the bytes of the signature. This is necessary for RFC3161 timestamp verification since the timestamp signature is generated over the bytes of the attestation signature.


ejahnGithub (approved):
lgtm

PR #923, opened 2023-12-30
fix ASN.1 date parsing w/ milliseconds

bdehamer:
Updates the ASN.1 parsing support in the `@sigstore/core` package to properly handle date/time strings which include millisecond-level precision.


ejahnGithub (approved):
lg

PR #931, opened 2024-01-04
collect timestamps from bundle

bdehamer:
Lays the foundation for RFC3161 timestamp verification in the `@sigstore/verify` package.

Updates the `SignedEntity` helper to extract any RFC3161 timestamps which may be present in the Sigstore bundle. At this point, it doesn't do any verification of the timestamps, but makes them available so they can be processed in a subsequent PR.


ejahnGithub (approved):
lgtm

PR #932, opened 2024-01-04
verify RFC3161 timestamps

bdehamer:
Updates the `Verifier` in `@sigstore/verify` to perform full verification of any RFC3161 timestamps present in the Sigstore bundle against the trusted root.


ejahnGithub (approved):
lgtm!

Issue #398, opened 2023-04-10
Embed TUF targets to minimize downloads


haydentherapper:
**Description**

For the Golang library, we embed a copy of the [target files](https://github.com/sigstore/sigstore/tree/main/pkg/tuf/repository/targets), and check if the target file matches the target metadata [here](https://github.com/sigstore/sigstore/blob/main/pkg/tuf/client.go#L522-L527) (a necessary check since the target file may have been updated and would need to be downloaded). Would it be possible to do the same for sigstore-js?

bdehamer:
Yeah, love this idea!

PR #941, opened 2024-01-10
support for multiple TUF repo caches

bdehamer:
Makes a number of changes to the way that the local TUF cache is managed

## Support for multiple repositories
The directory structure of the local TUF cache has been re-organized to support working with multiple TUF repositories without the need to purge the cache.

Previously, the local TUF cache was structured as follows:
```
.
├── remote.json
├── root.json
├── snapshot.json
├── targets
│   └── trusted_root.json
├── targets.json
└── timestamp.json
```

With this layout we could only cache the metadata/targets for a single TUF repository at a time. Working with a different TUF repo would require that the existing content be overwritten.

The new cache layout is as follows:
```
.
└── tuf-repo-cdn.sigstore.dev
    ├── root.json
    ├── snapshot.json
    ├── targets
    │   └── trusted_root.json
    ├── targets.json
    └── timestamp.json
```
With this scheme the metadata/targets are organized under a directory which identifies the remote repository from which those files were originally loaded.

Note that the old implementation used the `remote.json` file to identify the remote repository which was currently loaded into the cache. This file is no longer needed as the hostname of the remote repository is part of the directory structure.

## Cache seed refactoring

To simplify set-up for the end-user, `sigstore-js` comes bundled with the current `root.json` for the Sigstore public-good TUF repository. This root metadata file is necessary to bootstrap any interaction with the Sigstore TUF repository.

Previously, the Sigstore public-good `root.json` was packaged in the `@sigstore/tuf` package as a static asset (an extra file that was added to the package at build time). This scheme of bundling non-code assets into the package causes some problems for projects which are using [ncc](https://github.com/vercel/ncc) to compile `sigstore-js` into a single JavaScript file (commonly done when [building GitHub Actions](https://github.com/vercel/ncc)).

To ensure compatibility with ncc, the seed `root.json` has been moved to a constant read from a TypeScript file. To ensure that we maintain byte-level compatibility with the source `root.json` the contents are base64-encoded. This will add some complexity to the process of keeping this file up-to-date (it's no longer sufficient to simply copy the file from the remote repository), but should eliminate any of the issues we've experienced when trying to make the library work with ncc.

## Embed TUF targets

Embeds TUF targets are part of the cache seed data. Takes advantage of the fact that targets like `trusted_root.json` don't change very often and allows clients to use them without needing to download them from the remote TUF repository. The targets are still verified agains the other TUF metadata but, assuming everything is still valid, it eliminatesa fetch from the remote repository (closes https://github.com/sigstore/sigstore-js/issues/398).


ejahnGithub (approved):
lgtm

PR #944, opened 2024-01-10
expose forceCache option in tuf package

bdehamer:
Updates the `@sigstore/tuf` package with the `forceCache` option which was [added](https://github.com/theupdateframework/tuf-js/pull/575) to `tuf-js`.

To help disambiguate things, the `force` option was renamed to `forceInit`. The `forceInit` flag is the opposite of `forceCache` in that it will force a complete reset of the local TUF cache regardless of its current state. `forceInit` takes precedence over `forceCache`.

We will continue to support the old `force` flag for some time, but the preferred named is now `forceInit`.

### DIFF
--- a/.changeset/new-birds-hide.md
+++ b/.changeset/new-birds-hide.md
@@ -1,0 +1,5 @@
+---
+"@sigstore/tuf": minor
+---
+
+Expose `forceCache` option for TUF client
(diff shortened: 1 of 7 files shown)


ejahnGithub (approved):
lg

PR #948, opened 2024-01-11
move TUF seeds into JSON file

bdehamer:
Moves the TUF repo seed values (root.json and various targets) from a TypeScript source file into a JSON file.

I realized that having these values in a .ts source file was going to make it incredibly difficult to perform any sort of automated updates for the seed values.


ejahnGithub (approved):
lg

PR #959, opened 2024-01-12
Add workflow to check for TUF root updates

bdehamer:
Adds a new workflow which will periodically look to see if there are any updates to the Sigstore public-good TUF root metadata or targets. If any updates are detected, a PR will be opened which updates the TUF seeds distributed as part of the `@sigstore/tuf` package.

Example workflow run: https://github.com/sigstore/sigstore-js/actions/runs/7505165562/job/20433839377

Example PR: https://github.com/sigstore/sigstore-js/pull/958


ejahnGithub (approved):
this is much better to use gh cli!

PR #964, opened 2024-01-15
reduce macos runner size

bdehamer:
Swaps `@swc/jest` for `ts-jest` as the TypeScript compiler when running tests.

This decreases the memory overhead significantly (while also increasing the the overall speed of the test suite).

One side-effect of the new compiler is that the code coverage metrics are calculated a bit differently. In order to maintain our 100% coverage metric I had to rejigger some of the exports, add a new test, and mark a few files as ignored completely.


ejahnGithub (approved):
Nice. Learned a new js test compile!

PR #999, opened 2024-02-08
handle rekor checkpoints w/o timestamps

bdehamer:
Updates tlog verification logic to properly handle checkpoint values which do NOT include timestamps. This data was never being used, but the verifier would throw an error if it was missing.

related: https://github.com/sigstore/rekor/pull/1888


ejahnGithub (approved):
lg!

PR #1002, opened 2024-02-09
use npm 10 when releasing package

bdehamer:
Ensure that we're using at least `npm@10` when publishing packages so we get the most current provenance attestations.


ejahnGithub (approved):
Lgtm!

PR #1011, opened 2024-02-13
update to @sigstore/protobuf-specs@0.3.0

bdehamer:
Add support for the new v0.3 bundle version (see https://github.com/sigstore/protobuf-specs/pull/213).

This change will ensure that we can verify v0.3 bundles, but does not yet generate v0.3 bundles.


ejahnGithub (approved):
i think it looks good 👍

PR #1072, opened 2024-03-15
handle RSA keys in trusted_root.json

bdehamer:
Fixes a bug in `@sigstore/verify` which causes an error if the `trusted_root.json` file contains an RSA key using the "pkcs1" encoding.


ejahnGithub (approved):
looks good

PR #1093, opened 2024-03-27
support for building v0.3 bundles

bdehamer:
Updates the `toMessageSignatureBundle` and `toDSSEBundle` utility functions to accept a new `singleCertificate` option which can be used to trigger the generation of bundles which use the the `certificate` field in the verification material instead of the `x509CertificateChain` field. When this option is set to `true` the functions will construct a bundle according to the v0.3 bundle specs (and use the v0.3 media type).

If no value is specified or it is set to `false`, the default will be to create v0.2 bundles.

See: https://github.com/sigstore/protobuf-specs/pull/279


bdehamer:
In draft, pending the release of the `@sigstore/protobuf-specs@v0.3.1`

ejahnGithub (approved):

PR #1148, opened 2024-05-08
fix retry bug when calling external apis

bdehamer:
To compensate for the fact that `make-fetch-happen` doesn't support retries for "POST" requests, this change introduces a wrapper around the fetcher which will retry requests to Fulcio/Rekor/TSA for network errors or for HTTP status codes of 408, 429, or 5xx.


ejahnGithub (approved):
lgtm

PR #1162, opened 2024-05-14
accept docker media types

bdehamer:
When an image is pushed to a registry, the tagged manifest may use any of a number of different content types:

```
application/vnd.oci.image.index.v1+json
application/vnd.oci.image.manifest.v1+json
application/vnd.docker.distribution.manifest.v2+json
application/vnd.docker.distribution.manifest.list.v2+json
```

When the OCI client in this package goes to verify the image digest, it needs to explicitly identify the different content types that it will accept. If the content type which was used to push the image is not present in the `Accept` header of the request, the repository will return a 404 -- making it appear it seem that the image does not exist.

The change here is simply to expand the list of media types sent in the `Accept` header to encompass ALL of the types we may encounter in the wild -- ensuring that we can resolve the image regardless of its associated media type.

See: https://github.com/actions/attest-build-provenance/issues/73


ejahnGithub (approved):
lgtm! should we also contain these 2 
- application/vnd.docker.image.rootfs.diff.tar.gzip
- application/vnd.docker.container.image.v1+json

https://github.com/opencontainers/image-spec/blob/dd33f727e2faea07432ef6f06d6f9afe73f3f519/media-types.md?plain=1#L60-L78

bdehamer:
@ejahnGithub 

>gtm! should we also contain these 2
>  * application/vnd.docker.image.rootfs.diff.tar.gzip
>  * application/vnd.docker.container.image.v1+json

Those media types may be used when retrieving blobs, but would never appear in response to the `/manifests` API so are not relevant here.

PR #1183, opened 2024-06-11
Detect registries with referrers API support

bdehamer:
#### Summary
Updates the `attachArtifactToImage` function in the `@sigstore/oci` package to better detect registries which support the OCI referrers API.

Previously we were depending on the presence of the [`OCI-Subject`](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests-with-subject) header in response to uploading the artifact manifest to determine if the registry supported the referrers API. This was not a reliable method as some registries (AWS ECR) will return this header even when they do NOT support the referrers API.

As a fix, we're now pinging the referrers API directly to see if we get a 200 response. If we do, we can be confident that the registry supports the referrers API.

### DIFF
--- a/.changeset/chilled-frogs-sort.md
+++ b/.changeset/chilled-frogs-sort.md
@@ -1,0 +1,5 @@
+---
+'@sigstore/oci': patch
+---
+
+Fix bug when detecting support for the referrers API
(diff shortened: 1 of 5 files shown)


packages/cli/att.json
ejahnGithub, line 1: i am not sure why we are adding bundle here..?

packages/cli/multistmt
ejahnGithub, line 1: i am not sure why we are adding bundle in this pr..?

packages/cli/stmt
ejahnGithub, line 1: i am not sure why we are adding bundle in this pr..?

ejahnGithub (approved):
overall lgtm. left a few comments

ejahnGithub (approved):

PR #1185, opened 2024-06-12
remove extraneous field from referrers index

bdehamer:
#### Summary
Fixes a bug which sometimes include an extra `subjectDigest` field in the referrers index when it is uploaded to the OCI registry.

The fix is to remove the `subjectDigest` field from the artifact descriptor before it is used to populate the referrers index.


ejahnGithub (approved):
lgtm!

PR #1187, opened 2024-06-13
bug fix for referrers api detection

bdehamer:
#### Summary
Updates the logic used to detect whether an OCI registry supports the referrers API. It appears that some registries will return a false 200 response when hitting the `/referrers` endpoint even though the referrers API is not supported.

To ensure that we are correctly detecting support for the referrers API, this PR updates the logic to check for the presence of the `subjectDigest` in the artifact descriptor before checking the response from the `/referrers` endpoint.

### DIFF
--- a/.changeset/lovely-numbers-run.md
+++ b/.changeset/lovely-numbers-run.md
@@ -1,0 +1,5 @@
+---
+'@sigstore/oci': patch
+---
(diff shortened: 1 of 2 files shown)


ejahnGithub (approved):
lg

PR #1267, opened 2024-10-09
default to generating v0.3 bundles

bdehamer:
#### Summary
Update the `toDSSEBundle` and `toMessageSignatureBundle` functions in the `bundle` package to generate Sigstore v0.3 bundles by default.

Previously, these functions would generate v0.2 bundles by default and you could force v0.3 by setting `singleCertificate: true`. Now we will default to v0.3 and you can force v0.2 by setting `certificateChain` to true.

This is a breaking change to the public interface and will result in a major version bump.

### DIFF
--- a/.changeset/four-suits-sleep.md
+++ b/.changeset/four-suits-sleep.md
@@ -1,0 +1,5 @@
+---
+'@sigstore/bundle': major
+---
+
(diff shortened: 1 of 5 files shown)


ejahnGithub (approved):
look good

PR #1269, opened 2024-10-09
default to signing v0.3 bundles

bdehamer:
#### Summary
Update the `DSSEBundleBuilder` class in the `sign` package to generate Sigstore v0.3 bundles by default.

Previously, this class would generate v0.2 bundles by default and you could force v0.3 by setting `singleCertificate: true`. Now we will default to v0.3 and you can force v0.2 by setting `certificateChain` to true.

This is a breaking change to the public interface and will result in a major version bump.

### DIFF
--- a/packages/cli/src/commands/attest.ts
+++ b/packages/cli/src/commands/attest.ts
@@ -191,5 +191,5 @@ const initBundleBuilder = (opts: SignOptions): BundleBuilder => {
 
   // Build the bundle with the singleCertificate option which will
   // trigger the creation of v0.3 DSSE bundles
-  return new DSSEBundleBuilder({ signer, witnesses, singleCertificate: true });
+  return new DSSEBundleBuilder({ signer, witnesses });
 };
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/attest.ts
```diff
@@ -191,5 +191,5 @@ const initBundleBuilder = (opts: SignOptions): BundleBuilder => {
 
   // Build the bundle with the singleCertificate option which will
   // trigger the creation of v0.3 DSSE bundles
-  return new DSSEBundleBuilder({ signer, witnesses, singleCertificate: true });
+  return new DSSEBundleBuilder({ signer, witnesses });
```
bdehamer, line 194: Arg no longer needed since this is the new default.

packages/client/src/config.ts
```diff
@@ -93,7 +93,10 @@ export function createBundleBuilder(
     case 'messageSignature':
       return new MessageSignatureBundleBuilder(bundlerOptions);
     case 'dsseEnvelope':
-      return new DSSEBundleBuilder(bundlerOptions);
+      return new DSSEBundleBuilder({
+        ...bundlerOptions,
+        certificateChain: true,
+      });
```
bdehamer, line 99: Will continue to generate v0.2 bundles by default in the `client` package (for now).

ejahnGithub (approved):
Lgtm

PR #1270, opened 2024-10-09
default to generating dsse rekor entries

bdehamer:
#### Summary
Updates `RekorWitness` so that it generates "dsse" entries instead of "intoto" entries by default. The user can still request an "intoto" entry by explicitly setting the `entryType` field.

This is a breaking API change and will result in a major version bump.

### DIFF
--- a/packages/client/src/config.ts
+++ b/packages/client/src/config.ts
@@ -170,6 +170,7 @@ function initWitnesses(options: SignOptions): Witness[] {
     witnesses.push(
       new RekorWitness({
         rekorBaseURL: options.rekorURL,
+        entryType: 'intoto',
         fetchOnConflict: false,
         retry: options.retry ?? DEFAULT_RETRY,
         timeout: options.timeout ?? DEFAULT_TIMEOUT,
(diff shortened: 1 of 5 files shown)


packages/client/src/config.ts
```diff
@@ -167,6 +167,7 @@ function initWitnesses(options: SignOptions): Witness[] {
     witnesses.push(
       new RekorWitness({
         rekorBaseURL: options.rekorURL,
+        entryType: 'intoto',
```
bdehamer, line 170: Maintaining compatibility in the `client` package my explicitly requesting the "intoto" entry type.

packages/sign/src/witness/tlog/entry.ts
```diff
@@ -28,15 +28,15 @@ export function toProposedEntry(
   content: SignatureBundle,
   publicKey: string,
   // TODO: Remove this parameter once have completely switched to 'dsse' entries
-  entryType: 'dsse' | 'intoto' = 'intoto'
+  entryType: 'dsse' | 'intoto' = 'dsse'
```
bdehamer, line 31: Switching the default from "intoto" to "dsse"

ejahnGithub (approved):

PR #1271, opened 2024-10-09
remove `hash` function in core package

bdehamer:
#### Summary
Remove the `hash` function in the `core` package. Change all uses of `hash` to the new `digest` function.

This is a breaking change and will result in a major version bump.


ejahnGithub (approved):
lgtm

PR #1327, opened 2024-12-04
Fix verify command in conformance CLI

bdehamer:
The `verify` command in the conformance suite driver was never fully implemented. This PR adds support for the `--trusted-root` flag and removes the `xfail` config for the tests we were previously skipping.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -29,7 +29,6 @@ jobs:
     - uses: sigstore/sigstore-conformance@6bd1c54e236c9517da56f7344ad16cc00439fe19 # v0.0.13
       with:
         entrypoint: ${{ github.workspace }}/packages/conformance/bin/run
-        xfail: "test_verify_with_trust_root"
 
   conformance-staging:
     name: Conformance Test (Staging)
--- a/packages/conformance/src/commands/verify-bundle.ts
+++ b/packages/conformance/src/commands/verify-bundle.ts
@@ -1,21 +1,11 @@
 import { Args, Command, Flags } from '@oclif/core';
 import { Bundle, bundleFromJSON, MessageSignature } from '@sigstore/bundle';
-import { TrustedRoot } from '@sigstore/protobuf-specs';
-import * as tuf from '@sigstore/tuf';
-import {
-  SignedEntity,
-  toSignedEntity,
-  toTrustMaterial,
-  TrustMaterial,
-  Verifier,
-} from '@sigstore/verify';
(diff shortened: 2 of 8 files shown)


packages/conformance/src/commands/verify-bundle.ts
```diff
@@ -78,32 +68,6 @@ export default class VerifyBundle extends Command {
...
-}
-
-// Initialize TrustMaterial from a file
-async function trustMaterialFromPath(path: string): Promise<TrustMaterial> {
-  const trustedRoot = await fs
-    .readFile(path)
-    .then((data) => JSON.parse(data.toString()));
-  return toTrustMaterial(TrustedRoot.fromJSON(trustedRoot));
-}
-
```
bdehamer, line 106: We need these functions for both `verify` and `verify-bundle` now so they're being extracted into a separate file.

packages/conformance/src/commands/verify.ts
```diff
@@ -25,6 +24,10 @@ export default class Verify extends Command {
       description: 'the expected OIDC issuer for the signing certificate',
       required: true,
     }),
+    'trusted-root': Flags.string({
+      description: 'path to trusted root',
+      required: false,
+    }),
```
bdehamer, line 30: Implement missing flag

packages/conformance/src/commands/verify.ts
```diff
@@ -42,6 +45,18 @@ export default class Verify extends Command {
...
 
+    const trustedRootPath = flags['trusted-root'];
+    const trustMaterial = trustedRootPath
+      ? await trustMaterialFromPath(trustedRootPath)
+      : await trustMaterialFromTUF(flags['staging']);
+
+    const verifier = new Verifier(trustMaterial, {
+      tlogThreshold: 0,
+      tsaThreshold: 0,
+    });
```
bdehamer, line 56: To use a custom trusted-root, we need to instantiate the `Verifier` directly instead of relying on the default verify function implemented in the `client` package.

.github/workflows/conformance.yml
```diff
@@ -29,7 +29,6 @@ jobs:
     - uses: sigstore/sigstore-conformance@6bd1c54e236c9517da56f7344ad16cc00439fe19 # v0.0.13
       with:
         entrypoint: ${{ github.workspace }}/packages/conformance/bin/run
-        xfail: "test_verify_with_trust_root"
```
bdehamer, line 32: We no longer need to skip this test now that we support all the flags on the `verify` command.

ejahnGithub (approved):
lgtm

PR #1388, opened 2025-03-10
Fix selection of CAs to use for verification

bdehamer:
#### Summary
Updates the logic used to select which CAs should be used to verify the Fulcio-issued signing certificate. Previously, set of CAs listed in the trusted root were filtered by comparing the not-before/not-after dates of the signing certificate to the published start/end dates of the CA chains. While this works fine in practice, it would be more accurate to use the verified timestamp of the signature to select the CAs to use for verification.

Under the previous approach there is a small chance that a CA could be incorrectly filtered from the list of trusted CAs if the publication of the CA chain and the issuance of the signing certificate happen close together. The end result would be an attestation that fails verification when it should have passed.


ejahnGithub (approved):
lgtm!

PR #1490, opened 2025-09-28
Return verification

### LINKED ISSUE #1489: verify behavior is not consistent with `sigstore-go`
**Description**

When using `verify` from `import { verify } from 'sigstore'` I noticed that the return value was not consistent with the `sigstore-go` implementation consumed in `cosign verify-blob`

- **Expected:** `Signer`
- **Actual:** `undefined`

**Version**

`sigstore@4.0.0`

indexzero:
Howdy all 👋 looking forward to contributing to the project. Many thanks to @wlynch and @hectorj2f for helping me ramp up on the codebases 🫶 

#### Summary

> Fixes #1489

This PR updates the behavior of `verify` to be consistent with the `sigstore-go` implementation. This `result` is returned to but not consumed by `cosign` so the corresponding `packages/cli` functionality was not updated as part of this PR to maximize cohesion

- Existing `verify` behavior: [sigstore/sigstore-go/pkg/verify/signed_entity.go#798](https://github.com/sigstore/sigstore-go/blob/201a35a/pkg/verify/signed_entity.go#L797)
- Existing `cosign verify-blob` behavior: [sigstore/cosign/cmd/cosign/cli/verify/verify_blob.go#192](https://github.com/sigstore/cosign/blob/8e3a787/cmd/cosign/cli/verify/verify_blob.go#L192)

#### Release Note

* `verify(bundle[, payload][, options])` now returns a `Signer` object containing the public key and identity information from the verification.

> n.b. the above was not marked as **BREAKING** as it – imho – _technically a bug_ since it is a patch to match the existing `sigstore-go` implementation. I defer to the maintainers on that of course, but wanted to share the thought process 🙏 

#### Documentation

Updated `README.md` to indicate the new return 

#### Additional Remarks

🤖 `claude` helped here, but all of the `packages/*` code itself was 100% old fashioned fingers on keyboard. I don't use `jest` often so it was helpful in adding the test coverage for this change. No offense will be taken if it is updated or removed.


hectorj2f:
LGTM

PR #1709, opened 2026-07-28
fix(sign): fetch existing Rekor entries by default

### LINKED ISSUE #1708: Recover Rekor duplicate-entry conflicts by default
## Summary

npm provenance publishing can fail with:

```text
TLOG_CREATE_ENTRY_ERROR error creating tlog entry - (409) an equivalent entry already exists in the transparency log
```

This appears to be a retry-after-success path. Rekor accepts and commits the create-entry request, but the client times out, receives a transient network failure, or retries after a retryable response. The retry submits the same equivalent entry, Rekor correctly returns 409, and the signing flow treats that as fatal.

## Current behavior

`TLogClient` defaults `fetchOnConflict` to `false`:

```ts
this.fetchOnConflict = options.fetchOnConflict ?? false;
```

High-level `sigstore.attest()` / npm provenance callers usually do not set this option directly, so a 409 from an already-created Rekor entry becomes `TLOG_CREATE_ENTRY_ERROR` instead of using the existing recovery path.

## Expected behavior

When Rekor returns a 409 with a location for an equivalent existing entry, the default behavior should fetch the existing entry and continue. The existing entry is the entry the client needs in the Sigstore bundle, so this makes the create-entry operation idempotent from the caller's perspective while preserving provenance.

Callers that need fatal conflict behavior can still pass `fetchOnConflict: false` explicitly.

## Downstream reports

Similar downstream failures and analyses:

- SocialGouv/code-du-travail-numerique#7419: https://github.com/SocialGouv/code-du-travail-numerique/issues/7419
- apify/apify-shared-js#649: https://github.com/apify/apify-shared-js/issues/649
- ChainSafe/lodestar#9703: https://github.com/ChainSafe/lodestar/pull/9703

## Proposed fix

Default `TLogClient` to `fetchOnConflict: true`, while preserving explicit `fetchOnConflict: false` for callers that want conflicts to remain fatal.

lodekeeper:
## Summary

- default `TLogClient` to fetching existing Rekor entries on 409 conflicts
- keep explicit `fetchOnConflict: false` behavior unchanged
- add coverage for the default 409 recovery path

## Context

This fixes a retry-after-success failure mode seen by npm provenance publishers. If Rekor commits a create-entry request but the client retries after a timeout or transient retryable failure, the retry receives `409 an equivalent entry already exists`. The client already has recovery logic for that condition, but the default was `fetchOnConflict: false`, so high-level callers such as npm provenance publishing treated the benign duplicate as a fatal `TLOG_CREATE_ENTRY_ERROR`.

Related downstream reports:

- Closes sigstore/sigstore-js#1708
- SocialGouv/code-du-travail-numerique#7419
- apify/apify-shared-js#649
- ChainSafe/lodestar#9703

## Verification

- `npm run build`
- `npm test --workspace @sigstore/sign -- witness/tlog/client.test.ts`
- `npm run lint:check -- --quiet`

Generated with AI assistance.

### DIFF
--- a/.changeset/rekor-fetch-on-conflict.md
+++ b/.changeset/rekor-fetch-on-conflict.md
@@ -1,0 +1,5 @@
+---
+"@sigstore/sign": patch
+---
+
+Fetch existing Rekor entries by default when a create request returns a conflict.
--- a/packages/sign/src/__tests__/witness/tlog/client.test.ts
+++ b/packages/sign/src/__tests__/witness/tlog/client.test.ts
@@ -125,6 +125,23 @@ describe('TLogClient', () => {
           .reply(409, {}, { Location: `/api/v1/log/entries/${uuid}` });
       });
 
+      describe('by default', () => {
+        const subject = new TLogClient({ rekorBaseURL });
+
+        describe('when the fetch is successful', () => {
+          beforeEach(() => {
+            nock(rekorBaseURL)
+              .get(`/api/v1/log/entries/${uuid}`)
+              .reply(200, rekorEntry);
+          });
+
+          it('returns a tlog entry', async () => {
+            const entry = await subject.createEntry(proposedEntry);
+            expect(entry).toBeTruthy();
+          });
+        });
+      });
+
       describe('when fetchOnConflict is false', () => {
         const subject = new TLogClient({
           rekorBaseURL,
--- a/packages/sign/src/witness/tlog/client.ts
+++ b/packages/sign/src/witness/tlog/client.ts
@@ -40,7 +40,7 @@ export class TLogClient implements TLog {
   private fetchOnConflict: boolean;
 
   constructor(options: TLogClientOptions) {
-    this.fetchOnConflict = options.fetchOnConflict ?? false;
+    this.fetchOnConflict = options.fetchOnConflict ?? true;
     this.rekor = new Rekor({
       baseURL: options.rekorBaseURL,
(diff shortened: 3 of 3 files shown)


matthewkeil (approved):
LGTM!! Thanks @lodekeeper

lodekeeper:
Fixed the DCO-only blocker by amending the commit with the required `Signed-off-by` trailer and force-pushing with lease. New head: `58ad28df103558af1cbfb06952c90431d539318c`; DCO is green now.

matthewkeil:
@bdehamer curious what your thoughts are on this one.  There were a few issues across a bunch of repos but wondering if there was a reason this defaulted to `false` originally.

lodekeeper:
Some author context on the motivating bug (deferring to @bdehamer on the reason for the original `false` default):

We hit this as retry-after-success `409 TLOG_CREATE_ENTRY_ERROR` in npm provenance publishing. If a run uploads the Rekor entry and the job then fails and retries past that point, the retry re-signs identical content → Rekor 409s the duplicate, and with `fetchOnConflict: false` that surfaces as a hard error instead of resolving to the existing entry. For a content-deterministic build the retry then stays wedged until the source changes (a plain re-run can't recover). It broke ChainSafe/lodestar's nightly `publish-dev` provenance job — consistent with the "few issues across a bunch of repos" you mentioned. Defaulting to `true` makes the duplicate-upload path idempotent.

PR #1729, opened 2026-08-18
Add Buildkite OIDC identity provider

sj26:
## Why

Buildkite jobs can mint audience-scoped OIDC tokens through the native agent, but sigstore-js does not currently discover those credentials. This prevents packages such as npm from generating Sigstore provenance in Buildkite without manually provisioning a token.

We have proposed npm trusted publishing support for Buildkite (https://github.com/npm/cli/pull/9888) and would like to support [provenance](https://github.com/npm/cli/blob/latest/workspaces/libnpmpublish/lib/provenance.js) too, but that requires support here in sigstore-js first.

## What

Add Buildkite to the ambient CI identity provider using [`buildkite-agent oidc request-token`](https://buildkite.com/docs/agent/cli/reference/oidc#request-oidc-token), while preserving the existing provider fallback behavior when the agent fails or returns no token. Add focused success and failure coverage, document the authoritative Buildkite issuer and claims, and include a minor changeset for `@sigstore/sign`.


sj26:
(Force-pushed to add DCO.)

